PR #9, opened 2019-07-11
Added some s2i config for the lup-public build

SteveMHoward:


SteveMHoward:
Added some s2i config for the lup-public build

PR #67, opened 2022-01-19
Fixed the bug that prevented shapefiles from being displayed 

SteveMHoward:
Fixed a bug that was preventing shapefiles from being displayed on the front end. Also fixed a minor issue where they weren't being displayed on local dev environments (looking for localhost:4300 rather than localhost:3000).

### DIFF
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -8,7 +8,6 @@ import { ConfigService } from 'app/services/config.service';
 import { NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 import { ActivatedRoute, Router } from '@angular/router';
 import { Document } from 'app/models/document';
-import { CommentsResolver } from '../comments/comments-resolver.service';
 import * as _ from 'lodash';
 
 // need to import leaflet this way to include the shapefile->geojson plugin
@@ -56,8 +55,8 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
     });
     // The following items are loaded by a file that is only present on cluster builds.
     // Locally, this will be empty and local defaults will be used.
-    const remote_api_path = window.localStorage.getItem('from_public_server--remote_api_base_path');
-    this.pathAPI = (_.isEmpty(remote_api_path)) ? 'http://localhost:3000/api' : remote_api_path;
+    const remoteApiPath = window.localStorage.getItem('from_public_server--remote_api_base_path');
+    this.pathAPI = (_.isEmpty(remoteApiPath)) ? 'http://localhost:3000/api' : remoteApiPath;
   }
 
   ngAfterViewInit() {
@@ -160,7 +159,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
     this.map.scrollWheelZoom.disable();
 
     // draw project marker
-    if (this.shapefile[0] && this.shapefile[0].documentFileName && this.shapefile[0].documentFileName.length > 0) {
+    if (this.shapefile[0] && this.shapefile[0]._id && this.shapefile[0].documentFileName && this.shapefile[0].documentFileName.length > 0) {
       console.log('Shapefile', this.shapefile);
       const escapedName = encode(this.shapefile[0].documentFileName).replace(/\(/g, '%28').replace(/\)/g, '%29').replace(/\\/g, '_').replace(/\//g, '_').replace(/\%2F/g, '_');
       const shapeurl = this.pathAPI + '/document/' + this.shapefile[0]._id + '/fetch/' + escapedName;
@@ -173,7 +172,6 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
       }.bind(this), 500);
       shapefile.addTo(this.map);
     } else if (this.project) {
-      console.log('No shapefille!', this.shapefile[0]);
       const markerIconYellow = L.icon({
         iconUrl: 'assets/images/marker-icon-yellow.svg',
         iconSize: [36, 36],
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -8,6 +8,8 @@ import { ConfigService } from 'app/services/config.service';
 import { NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 import { ActivatedRoute, Router } from '@angular/router';
 import { Document } from 'app/models/document';
+import { CommentsResolver } from '../comments/comments-resolver.service';
+import * as _ from 'lodash';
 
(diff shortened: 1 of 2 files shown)


src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -166,6 +173,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
       }.bind(this), 500);
       shapefile.addTo(this.map);
     } else if (this.project) {
+      console.log('No shapefille!', this.shapefile[0]);
```
CR-gbc: is this a debugging item left in?

not sure on this project's policy of logging to console.

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -44,13 +47,17 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
     this.commentPeriod = this.project.commentPeriodForBanner;
     this.route.data.subscribe((res: any) => {
       if (res) {
-        if (res.shapefile && res.shapefile[0].data.meta && res.shapefile[0].data.meta.length > 0) {
-          this.shapefile = res.shapefile[0].data.searchResults;
+        if (res.documents && res.documents[0].data.meta && res.documents[0].data.meta.length > 0) {
+          this.shapefile = res.documents[0].data.searchResults;
```
ShawnTurple, line 50: You might want to consider using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining here, might clean up the code a bit

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -44,13 +47,17 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
...
         } else {
           this.shapefile = [];
         }
       }
     });
+    // The following items are loaded by a file that is only present on cluster builds.
+    // Locally, this will be empty and local defaults will be used.
+    const remote_api_path = window.localStorage.getItem('from_public_server--remote_api_base_path');
```
ShawnTurple: This is the WordPress php in you coming out try `const remoteApiPath =`

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -156,7 +163,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
     if (this.shapefile[0] && this.shapefile[0].documentFileName && this.shapefile[0].documentFileName.length > 0) {
```
ShawnTurple: again might want to consider The optional chaining operator (?.)

CR-gbc (commented):
i am not the console.log police, so other than the log, all seems well. I'll leave it up to you if you want to leave it in.

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -156,7 +163,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
...
       console.log('Shapefile', this.shapefile);
       const escapedName = encode(this.shapefile[0].documentFileName).replace(/\(/g, '%28').replace(/\)/g, '%29').replace(/\\/g, '_').replace(/\//g, '_').replace(/\%2F/g, '_');
-      const shapeurl = '/api/document/' + this.shapefile[0]._id + '/fetch/' + escapedName;
+      const shapeurl = this.pathAPI + '/document/' + this.shapefile[0]._id + '/fetch/' + escapedName;
```
ShawnTurple, line 165: are you sure there will always be a shapefile._id here...

src/app/project/project-details-tab/project-details-tab.component.ts
SteveMHoward: Good call! I removed others but missed that one!

CR-gbc:
you should have seen my eyes light up when i saw a @SteveMHoward commit!

src/app/project/project-details-tab/project-shapefile-resolver.service.ts
```diff
@@ -23,7 +23,7 @@ export class ShapeFileResolver implements Resolve<Observable<object>> {
       currentPage,
       pageSize,
       sortBy,
-      {},
+      { documentSource: 'SHAPEFILE' }, // only look for shapefiles so we are not needlessly parsing through images or other documents
```
ShawnTurple: Comments should start with caps and end with full stops (ie .?!)

ShawnTurple (changes requested):
As far as using the the optional chaining operator (?.) .. your call it would make code a bit more readable.
NW:
* Fix comment begins with cap, ends full stop.
* Naming conventions should be camelCase

SteveMHoward:
@ShawnTurple Good calls! Still got to shake of that rust a bit ;)

src/app/project/project-details-tab/project-details-tab.component.ts
SteveMHoward: Fixed in latest push

ShawnTurple (approved):

src/app/project/project-details-tab/project-details-tab.component.ts
SteveMHoward: Considered, but not willing to change at this time. Could be code impacts for the empty array that is set below vs whatever happens to the variable when const something?.something else fails.

src/app/project/project-details-tab/project-details-tab.component.ts
SteveMHoward: Fixed in latest push.

src/app/project/project-details-tab/project-details-tab.component.ts
SteveMHoward: Not implementing for now. Would have to redo the entire conditional function based on this change, which I feel is out of scope for a bug fix.

src/app/project/project-details-tab/project-details-tab.component.ts
SteveMHoward: Added a check for _id :)

src/app/project/project-details-tab/project-shapefile-resolver.service.ts
SteveMHoward: Fixed!

src/app/project/project-details-tab/project-details-tab.component.ts
Baelx: @SteveMHoward Actually, I don't think there's a downside to using optional chaining in this instance. The if statements will still evaluate the same. The difference is:

- Without optional chaining: Any one of those properties missing will cause a nasty error to be thrown saying that there is no such property. We just haven't run into this before. The if statement will _not_ default to the "else" but the code will error out.
- With optional chaining: If any property is missing, "undefined" will be output and the if statement logic will continue as normal.

So, TLDR, optional chaining will actually make it _more_ likely the if statement logic evaluates as expected.

PR #68, opened 2022-01-26
DESENG-54 Fix header padding, add new project fields

Baelx:
FYI this project has not been meeting the DoD for a long time. We're in the process now of getting it up to speed. Github Actions are coming next sprint and the process of implementing tests and linting will begin shortly after.


src/app/project/comments/comments.component.ts
```diff
@@ -101,13 +101,13 @@ export class CommentsComponent implements OnInit, OnDestroy {
           if (data.commentPeriod) {
             // To fix the issue where the last page is empty.
             this.commentPeriod = data.commentPeriod;
-            console.log('Comment Period info', this.commentPeriod);
+            const engagementLabel = this.project.engagementLabel ? this.project.engagementLabel : 'Public Comment Period';
```
CR-gbc, line 104: just curious, || doesn't work here?

CR-gbc (approved):

src/app/project/comments/comments.component.ts
Baelx: Ahh, good point. That would have been less verbose. I think I had meant to switch this to the OR operator but never came back to it.

PR #69, opened 2022-02-10
Feature/deseng55

Baelx:
Should be the last PR in this project for this ticket.

[-] Wrote a test or tests for code you added.
[x] Linted the files you touched.
[x] Wrote a doc block for each method you added.
[-] Add documentation to the README


src/app/models/project.ts
```diff
@@ -96,37 +98,22 @@ export class Project {
...
+  }
+}
 
-    // // copy documents
-    // if (obj && obj.documents) {
-    //   for (const doc of obj.documents) {
-    //     this.documents.push(doc);
-    //   }
-    // }
+export interface ProjectLogo {
```
ASpiteri-BCGov, line 110: Maybe consider creating a types file where all types are defined and exported.

ASpiteri-BCGov (approved):

src/app/models/project.ts
Baelx: Thanks for the suggestion. I definitely agree we should be keeping these types as organized as possible.

This project.ts file - and all files within the models directory - contains the types for all of the "data structures" in the app. Just based on how we've organized types in this app, I think any "project" type should be fine to go within project.ts

src/app/project/project.component.scss
```diff
@@ -30,6 +30,23 @@
...
+        padding: 0;
+        list-style-type: none;
+        display: flex;
+
+        li {
+          margin-right: 1rem;
+
+          a {
+            img {
+              max-height: 90px;
```
SteveMHoward, line 44: Have you verified this works as expected in Chrome, Firefox and Safari? 

I am not sure if it has changed, but previously there was an issue using just max-height without defining a height in one of the major non-IE browsers (I think it was FF but don't recall) - the height would get returned as 0px.

Baelx: Interesting! I haven't run across that bug before. Just for good measure, I had a look at this in those browsers and everything it tickety boo

SteveMHoward (approved):

PR #79, opened 2022-05-25
DESENG-4 Polygons render on the main map

Baelx:


src/app/models/project.ts
```diff
@@ -73,6 +74,7 @@ export class Project {
     this.region = obj && obj.region || null;
     this.projectDirector = obj && obj.projectDirector || null;
     this.agreements = obj && obj.agreements || null;
+    this.shapefiles = obj && obj.shapefiles || [];
```
ShawnTurple, line 77: what if obj.shapefiles is not an array, but something else will your code handle this, or will it always be an array, if it makes it to this point?

src/app/projects/projects.component.ts
```diff
@@ -114,20 +114,42 @@ export class ProjectsComponent implements OnInit, OnDestroy {
...
+                            shapefileId: shapefile._id,
+                            documentFileName: shapefile.documentFileName
+                          })
+                        }
+                      })
+                    })
+                  }
+
+                  this.allApps = _.concat(this.allApps, projects);
+                  console.log('all projects', projects)
```
ShawnTurple: is this suppose to be here?

src/app/projects/projects.component.ts
```diff
@@ -114,20 +114,42 @@ export class ProjectsComponent implements OnInit, OnDestroy {
...
+              this.documentService.getAll("SHAPEFILE")
+              .takeUntil(this.ngUnsubscribe)
+              .subscribe(shapefiles => {
+                  // "Attach" a shapefile to its project.
+                  if (shapefiles) {
+                    projects.forEach(project => {
+                      // project.shapefiles = [];
+                      shapefiles.forEach(shapefile => {
+                        if (project._id === shapefile.project) {
+                          console.log("how many times")
```
ShawnTurple: another console.log left in, guessing, not suppose to be here?

kshapka-bcgov (approved):
Love to see all the code cleanup. Looks good!

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -301,33 +259,49 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
...
 
     // draw added apps
     addedApps.forEach(app => {
-      // add marker
-      if (app.centroid.length === 2) { // safety check
-        const title = `${app.name}\n`
+      // If there is a shapefile for one of the projects, display it instead of a pin.
+      if (app.shapefiles && app.shapefiles.length > 0) {
+        app.shapefiles.forEach(projectShapefile => {
+          const escapedName = encode(projectShapefile.documentFileName).replace(/\(/g, '%28').replace(/\)/g, '%29').replace(/\\/g, '_').replace(/\//g, '_').replace(/\%2F/g, '_');
```
ShawnTurple, line 286: just wondering how you can have a const in a foreach loop

CR-gbc (approved):

src/app/projects/projlist-map/projlist-map.component.ts
ShawnTurple, line 286: It feels like this could be done a more efficient way using regex

src/app/services/api.ts
```diff
@@ -630,6 +630,41 @@ export class ApiService {
     return this.http.get<Document[]>(`${this.apiPath}/${queryString}`, {});
   }
 
+  /**
+   * Get all documents.
```
ShawnTurple, line 634: could be a little more descriptive... get all documents for what?

ShawnTurple (changes requested):
Remove console.log

src/app/models/project.ts
Baelx: Good question - it should always be an array and we want to force that here.

src/app/projects/projects.component.ts
Baelx: It's not, sorry about that. Will update with them removed

src/app/projects/projlist-map/projlist-map.component.ts
Baelx: Here, we're just saving a complicated operation of converting a string to a URI and saving that for later use. We need to do this for each "app" or in this case, project.

ShawnTurple: again, const should only be declared once.  If its working, great, curious to know what the linter might pick up.

Baelx: So like a single "replace" call that captures each pattern? We'd still need to replace a pattern with a specific thing(for example, "\\" needs to get replaced with "_"), so I don't know how we could do that more succinctly.

Baelx: Oh, I see what you're saying. So we're passing a function to forEach which is projectShapefile => { ... }. Within that function, we're declaring a variable which is block-scoped. In a forEach loop, the function is called for each item in the array - meaning each function will have its own scope. The const shouldn't exist from one loop iteration to the other.

This code is working as expected.

ShawnTurple (approved):

src/app/services/api.ts
Baelx: I do agree that we should write descriptive doc blocks but in this case, I'm not sure how. This call is basically the same as the PMO app's getAll Contracts call: https://github.com/bcgov/gdx-agreements-tracker/blob/development/backend/src/controllers/contacts.js#L29

This is a common data type in the app and we just need to get all of them.

PR #81, opened 2022-06-18
DESENG-4 Fix popup not appearing on shapefile

Baelx:
Had to fix an issue with popups not appearing over shapefiles on map.


kshapka-bcgov (approved):
Looks good, once the failed check gets sorted out.

PR #84, opened 2023-02-18
Deseng-53 Site style rebrand

Baelx:
This work was originally done by Steve Howard in 2022. We're now merging it into the dev branch for testing and deployment to prod.

See: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-270


src/app/header/header.component.scss
```diff
@@ -256,20 +256,20 @@
...
   /*background-image: url("/assets/images/bcgov-header-vert-SM.png");*/
-  background-image: url("/assets/images/BCID_H_rgb_rev.png");
-  background-position: 8px center;
-  background-size: contain;
-  background-repeat: no-repeat;
-  white-space: normal;
+  //background-image: url("/assets/images/BCID_H_rgb_rev.png");
+  //background-position: 8px center;
+  //background-size: contain;
+ //background-repeat: no-repeat;
```
robrien-bcgov: wouldn't it be better to just delete these?

src/assets/styles/layout/layout.scss
```diff
@@ -42,6 +29,10 @@ main {
     line-height: 1.4rem;
     font-size: 0.9375rem;
   }
+  
+  a {
+    text-decoration: underline;
+  }
```
robrien-bcgov, line 35: where is the text decoration set to 'none' for anchor tags?

src/assets/styles/themes/lup-rebrand.scss
```diff
@@ -0,0 +1,125 @@
...
+// Map Components.
+$map-bg:                        #445a71;
+
+// Map Popup.
+$map-popup-color:               #fff;
+$map-popup-bg:                  #244278;
+$map-popup-footer-bg:           $primary;
+$map-popup-header-bg:           $primary;
+$map-popup-font-size:           0.9375rem; // 15 px
+$map-popup-menu-active-bg:      #5091cd;
```
robrien-bcgov, line 118: wow that's a lot of variables, and nicely organized.

robrien-bcgov (approved):
looks good but I don't understand why there is so much commented-out CSS

src/app/app.component.scss
```diff
@@ -8,7 +8,7 @@
   /*display: -webkit-box;*/
 
   min-height: 100vh;
-  background: #036;
+  background: $primary;
 }
```
jareth-whitney, line 12: A more descriptive colour variable name might be more suitable here.

src/app/footer/footer.component.scss
```diff
@@ -59,6 +60,7 @@
     &:focus,
     &:hover {
       color: $gold;
+      text-decoration: none;
     }
```
jareth-whitney, line 64: The nesting is a little odd here, but perhaps this repo is just set to a two space tab

src/app/header/header.component.scss
```diff
@@ -119,7 +119,7 @@
       z-index: 1001; // +1 to keep above dropdown menus
       padding: 0.875rem 1rem;
       border-top: 1px solid $mn-border-color;
-      color: #fcba18;
+      color: $header-font-color;
```
jareth-whitney, line 120: Just another little nudge to make these colour variables more comprehensible, possibly by including the colour name in the variable name.

src/app/project/comments/comments.component.scss
```diff
@@ -22,7 +22,8 @@
   background-color: #001330;
 
   .hero-banner__container {
```
jareth-whitney, line 18: Possibly replace literal hex values with variables if there's time (line 22 for example)

src/assets/styles/base/base.scss
```diff
@@ -1,4 +1,5 @@
 @import "~@angular/material/prebuilt-themes/indigo-pink.css"; // see https://material.angular.io/guide/theming
 @import "~assets/styles/components/ckeditor.scss";
 @import "~assets/styles/abstracts/mixins.scss";
-@import "~assets/styles/themes/default.scss";
+//@import "~assets/styles/themes/default.scss";
```
jareth-whitney: Is the commented-out import still here because it will be needed later?

src/assets/styles/components/buttons.scss
```diff
@@ -13,14 +13,13 @@
     &:active,
     &:focus,
     &:hover {
-      border-color: $gold;
-      background: darken($gold, 8%);
+      border-color: #fff;
+      background: darken($primary, 8%);
```
jareth-whitney, line 17: I didn't know about darken, cool!

src/assets/styles/components/hero-banner.scss
```diff
@@ -8,6 +8,19 @@
   background-size: 350%;
   box-shadow: 0 0 4rem #00264d inset;
 
+  /*.hero-banner {
+  background-size: cover;
```
jareth-whitney: Is this commented-out CSS still needed here?

src/assets/styles/themes/lup-rebrand.scss
```diff
@@ -0,0 +1,125 @@
+/** Rebranding work done for the Land Use Planning platform as outlined in DESENG-53. **/
+
+// Main Colors.
```
jareth-whitney, line 3: Black is given a variable so many times here. I suppose this is necessary because we're doing a retrofit, and if we went back then we would want different values. If that's not the case, perhaps we could consolidate some of the black colour variables in a future ticket.

jareth-whitney (changes requested):
I think this ticket is a great initiative and I'm stoked that it will be going to production. There are a couple of small things here:

- Some colour variables could be given more readable names (that include the colour name)
- Some colours still use hex literals instead of variables
- There is some commented out code that may be intentional, but I left a comment in case it's not
- The colour black is given a variable many times, possibly because this is a retrofit and we want the variables separate in the future. If not, then we should consider consolidating some variables either now or in a future ticket.

Baelx:
> I think this ticket is a great initiative and I'm stoked that it will be going to production. There are a couple of small things here:
> 
>     * Some colour variables could be given more readable names (that include the colour name)
> 
>     * Some colours still use hex literals instead of variables
> 
>     * There is some commented out code that may be intentional, but I left a comment in case it's not
> 
>     * The colour black is given a variable many times, possibly because this is a retrofit and we want the variables separate in the future. If not, then we should consider consolidating some variables either now or in a future ticket.

Thanks for the thorough review @robrien-bcgov @jareth-whitney. 

- Since we generally don't block PRs for not using CSS variables, and because this is Steve's code, I won't do the variables piece. The ticket assumes the code is good to go and just needs to be merged
- I will remove all the commented out code because it adds nothing(won't be used later, not used as documentation) and increases clutter

src/assets/styles/layout/layout.scss
Baelx: I'm not sure here! Not my code originally but I would guess this declaration conflicted with one in another stylesheet.

src/assets/styles/themes/lup-rebrand.scss
Baelx: Great point here. If we have the cycles for a scss clean-up, we could definitely do a lot to reduce code debt

src/assets/styles/layout/layout.scss
robrien-bcgov: Hmm. it just looks weird to set text decoration to underline when that's the default. 

I'd be deleting this and hunting for where it's set otherwise and setting a more specific (class-based) selector instead, but maybe there's an unknown side-effect I haven't accounted for.

jareth-whitney (approved):
Thanks for taking out all of the commented-out code!

PR #87, opened 2023-06-15
Bugfix/deseng344 - Update Angular and Packages in LUP-public to Address Security Concerns

jareth-whitney:
- Updated Angular to version 10 and subsequently to version 11
- Updated Angular material to version 10 and subsequently to version 11
- Fixed any issues with deprecated code
- Updated packages that had security issues or needed updating for Angular 11: moment-timezone, ng-bootstrap, etc.


src/app/services/api.ts
```diff
@@ -131,14 +123,14 @@ export class ApiService {
...
-    if (queryModifier !== {}) {
+    if (Object.keys(queryModifier).length !== 0) {
       Object.keys(queryModifier).forEach(key => {
         queryModifier[key].split(',').map(item => {
           queryString += `&and[${key}]=${item}`;
         });
       });
     }
-    if (filter !== {}) {
+    if (Object.keys(queryModifier).length !== 0) {
```
robrien-bcgov, line 133: i'm a big fan of lodash:

`_.isEmpty(objectName)`

I use it all the time b/c it simplifies my code by taking care of those little bits of logic

src/app/services/api.ts
```diff
@@ -69,19 +65,15 @@ export class ApiService {
     const blob = await this.downloadResource(document._id);
     let filename = document.displayName;
     filename = encode(filename).replace(/\\/g, '_').replace(/\//g, '_');
```
robrien-bcgov, line 67: i'm guessing we don't support whatever the `this.isMS` thing is for due to newer browsers?

src/app/services/api.ts
```diff
@@ -131,14 +123,14 @@ export class ApiService {
     if (sortBy !== null) { queryString += `&sortBy=${sortBy}`; }
     if (secondarySort !== null) { queryString += `&sortBy=${secondarySort}`; }
     if (populate !== null) { queryString += `&populate=${populate}`; }
-    if (queryModifier !== {}) {
+    if (Object.keys(queryModifier).length !== 0) {
       Object.keys(queryModifier).forEach(key => {
```
robrien-bcgov, line 127: i wonder what would happen if you just got rid of the if conditional and let it iterate over Object.keys?

I mean, if there's no keys, it will stop iterating at the 0th key right?

not that it matters here, but I though it was interesting.

robrien-bcgov (approved):
seems legit

src/app/services/api.ts
jareth-whitney: I haven't gotten into lodash yet. That does look much simpler though, I should read up.

src/app/services/api.ts
jareth-whitney: Yes it's an Internet Explorer check and we've discontinued support for IE.

src/app/services/api.ts
jareth-whitney: I didn't know that it works that way, I thought it would throw an error. That's good to know though.

PR #94, opened 2023-06-16
bugfix/deseng344: Updated some rxjs imports, changed test.yml Node to v12

jareth-whitney:
- Updated some rxjs imports to follow new guidelines
- Changed node version in github test.yml to 12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 16
+          node-version: 12
       - name: Install packages
         run: npm install
       - name: Run ESLint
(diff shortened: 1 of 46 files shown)


.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 16
+          node-version: 12
```
mhaswell-bcgov, line 15: Just wondering about this. I'm guessing the production server is running on this node version?

mhaswell-bcgov (approved):

kshapka-bcgov (approved):
The words 'observable' and 'subject' no longer mean anything to me.

.github/workflows/test.yml
jareth-whitney: I followed the Angular compatibility guide. Previously, it would fail at package building on Github due to the node version. It's not that important now, but it will be once we have functioning linting/unit tests setup on Github.

PR #109, opened 2024-06-17
DESENG-641 Final content changes

Baelx:
Just static text changes, no logic.


NatSquared (approved):
Only content changes; looks good to me 👍

PR #113, opened 2024-07-29
Feature/deseng536 Redirect landuseplanning.gov.bc.ca to planninginpartnership.ca

Baelx:
- Remove commands that copy nginx config to container
- Create configmaps to store nginx config per-project
- Modify deployment configs to use mounted configmaps
- Fix local build issue by inputting proper package.json version and removing build step
- Remove commented-out nginx config


openshift/templates/nginx-runtime/nginx.conf.template
```diff
@@ -54,8 +37,9 @@ http {
     client_max_body_size 100m;
 
     server {
-        listen       8080 default_server;
-        server_name  localhost;
+        listen 8080;
+        server_name landuseplanning.gov.bc.ca;
+        rewrite ^/(.*)$ https://planninginpartnership.ca/$1 permanent;
```
NatSquared, line 43: Looks like a standard permanent redirect 👍

openshift/templates/nginx-runtime/nginx.conf.template
```diff
@@ -13,34 +15,15 @@ http {
...
-
     keepalive_timeout  65;
 
-    #gzip  on;
-
-    #real_ip module
- #   set_real_ip_from %RealIpFrom%;
-    #%AdditionalRealIpFromRules%
- #   real_ip_recursive on;
- #   real_ip_header    X-Forwarded-For;
```
NatSquared, line 42: Thanks for doing this cleanup!

openshift/templates/nginx-runtime/nginx.conf.template
```diff
@@ -1,3 +1,5 @@
+# This nginx server config now lives in ConfigMaps (nginx-conf) in the dev, test, prod projects respectively
```
NatSquared: Sounds good! Thanks for maintaining this copy; good to know we're using a more flexible system now

NatSquared (approved):
Looks good. This PR adds a redirect for requests from old URLs and a new site to be served at the new URL.

Baelx:
> Looks good. This PR adds a redirect for requests from old URLs and a new site to be served at the new URL.

Thanks Nat! It's technically not a new site, but just two different domains that point to the same site. For branding purposes, we don't want the landuseplanning URL persisting in the production site but for it to always redirect such that the address bar populates with planninginpartnership.ca

PR #116, opened 2024-11-13
feature/deseng742: Hid agreements header when no agreements available.

jareth-whitney:
- Hid agreements header when no agreements available

### DIFF
--- a/src/app/project/comments/comments.component.html
+++ b/src/app/project/comments/comments.component.html
@@ -44,8 +44,11 @@
                 <label>Partner First Nation(s)</label>
                 <p class="value">{{project?.partner || "-" }}</p>
               </div>
-              <div *ngIf="project?.agreements?.length > 0" class="agreements">
+              <div *ngIf="project?.agreements?.length > 0" class="agreements-label">
                 <label>Agreement(s)</label>
+              </div>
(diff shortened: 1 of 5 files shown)


src/app/project/comments/comments.component.html
```diff
@@ -44,13 +44,10 @@ <h3>
                 <label>Partner First Nation(s)</label>
                 <p class="value">{{project?.partner || "-" }}</p>
               </div>
-              <div *ngIf="projectFieldType(project?.agreements) ==='string'" class="agreements">
+              <div *ngIf="project?.agreements?.length > 0" class="agreements">
```
Baelx: Back in the day, projects only had a single agreement which was a string. We kept some logic in place in case that's still the case (but we should have left a comment for it).

We may want to retain that logic, just in case. The other option would be to log in to production LUP and make sure none of the projects are using a string as their agreement. We can chat further about how to do that if you want to go that route

Baelx (changes requested):
Everything is basically good to go but see my other comment about agreements potentially being a string. As such we may want to leave some logic in place for backwards compatibility.

Baelx (approved):
Thanks for the changes! Looks good :)

PR #118, opened 2024-11-23
feature/deseng746: Changed project list column from project phase to process type(s).

jareth-whitney:
- Changed project list column from 'Project Phase' to 'Process Type(s)'

### DIFF
--- a/src/app/projects/project-list/project-list.component.ts
+++ b/src/app/projects/project-list/project-list.component.ts
@@ -86,8 +86,8 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       width: 'col-4'
     },
     {
-      name: 'Phase',
-      value: 'projectPhase',
+      name: 'Process Type(s)',
+      value: 'projectTypes',
       width: 'col-4'
     }
   ];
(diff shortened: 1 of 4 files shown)


src/app/projects/project-list/project-list.component.ts
```diff
@@ -351,6 +351,12 @@ export class ProjectListComponent implements OnInit, OnDestroy {
     return overlappingDistrictsListString;
   }
 
+	stringifyProjectTypes(projectTypes) {
```
Baelx, line 354: You could also chain these array methods, just like jquery used to do!

```suggestion
	stringifyProjectTypes(projectTypes: ProjectType[]) {
	    // Filter by selected types, then get the string of their name, then join as a single string.
	    return projectTypes.filter(pt => true === pt.checked).map(pt => pt.name).join(', ') || '';
	    }
```

Baelx (approved):
There may be room to simplify some of the code but otherwise you're good to go. Thanks!

src/app/projects/project-list/project-list.component.ts
jareth-whitney: ok I'll try that approach next time

PR #119, opened 2024-11-26
feature/deseng747: Custom collection notice is now viewable in subscribe and contact modals when available.

jareth-whitney:
- Custom collection notice is now viewable in subscribe and contact modals when available.

### DIFF
--- a/src/app/models/project.ts
+++ b/src/app/models/project.ts
@@ -49,6 +49,7 @@ export class Project {
   activitiesAndUpdatesEnabled: boolean;
   contactFormEnabled: boolean;
   contactFormEmails: string[];
+  collectionNotice: String;
 
   // Permissions
   read: Array<String> = [];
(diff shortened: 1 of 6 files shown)


src/app/models/project.ts
```diff
@@ -49,6 +49,7 @@ export class Project {
   activitiesAndUpdatesEnabled: boolean;
   contactFormEnabled: boolean;
   contactFormEmails: string[];
+  collectionNotice: String;
```
Baelx:
```suggestion
  collectionNotice: string;
```

src/app/project/email-subscribe/email-subscribe.component.ts
```diff
@@ -35,6 +35,7 @@ export class EmailSubscribeComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.emailSubscribe = new EmailSubscribe();
     this.emailSubscribe.project = this.project._id;
+    console.log('project', this.project);
```
Baelx:
```suggestion
```

Baelx (changes requested):
Thanks! Looks good. Just as with the review of the "admin" app, please update the project model to use `string` for the collectionNotice. Also, please remove the console.log added

Baelx (approved):
Thanks! Approved

PR #123, opened 2024-12-11
Feature/deseng750: Adjusted northern default bounds, optimized shape file code

jareth-whitney:
- Adjusted default northern bounds
- optimized shape file code


Baelx (approved):
Looks fine!

PR #124, opened 2025-01-10
feature/deseng751: Combined external links and internal files in file lists

jareth-whitney:
- Combined external links and internal files in file lists
- Functionality supports sorting, pagination, and sections


src/app/models/externalLink.ts
```diff
@@ -0,0 +1,27 @@
+import * as _ from 'lodash';
```
Baelx: What is lodash used for here?

Baelx (approved):
Looks good! Thanks

PR #126, opened 2025-01-21
feature/deseng752: Added ability to filter projects on map by project type.

jareth-whitney:
- Added ability to filter projects on map by project type.

### DIFF
--- a/src/app/projects/projlist-filters/projlist-filters.component.scss
+++ b/src/app/projects/projlist-filters/projlist-filters.component.scss
@@ -372,14 +372,18 @@ $app-filters-font-lg: 0.9375rem;
 	padding-left: 0.5rem;
 }
 
-#project-type-filter-show-hide.hide {
-	display: none;
-}
-
 ::ng-deep .project-type-filter-checkbox .mat-checkbox-ripple .mat-ripple-element {
 	background-color: #ffd740 !important;
 }
 
 ::ng-deep .project-type-filter-checkbox .mat-checkbox-indeterminate .mat-checkbox-background, ::ng-deep .mat-checkbox-checked .mat-checkbox-background {
 	background-color: #ffd740 !important;
+}
+
+.filter-by-type {
+  font-size: 1.25rem;
+}
+
+.filter-collapse-button.collapsed {
+  rotate: 180deg;
 }
--- a/src/app/projects/projlist-filters/projlist-filters.component.scss
+++ b/src/app/projects/projlist-filters/projlist-filters.component.scss
@@ -1,16 +1,4 @@
 @import "~assets/styles/base/base.scss";
-// @import '~@angular/material/theming';
-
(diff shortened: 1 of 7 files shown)


src/app/projects/projlist-filters/projlist-filters.component.scss
```diff
@@ -1,4 +1,16 @@
...
+// $custom-theme: mat-bcgov-theme(
+//   (
+//     color: (
+//       primary: '000000',
+//       accent: 'ffd740',
+//       warn: 'ffd740',
+//     ),
+//   )
+// );
+// @include angular-material-theme($custom-theme);
```
Baelx: I'm going to just leave comments for myself to resolve some issues since Jareth is away. Commented-out code here

src/app/projects/projlist-filters/projlist-filters.component.ts
```diff
@@ -159,6 +172,37 @@ export class ProjlistFiltersComponent implements OnInit, OnChanges, OnDestroy {
...
+	}
+
+	/**
+	 * Shows or hides the project type filter. Especially useful for mobile viewports where it covers most of the screen.
+	 * 
+	 * @returns {void}
+	 */
+	public handleShowHideFilter(): void {
+		if (false === this.filtersOpen) {
+			this.elementRef.nativeElement.querySelector("#project-type-filter-show-hide").classList.remove('hide');
```
Baelx: Maybe there was something preventing it, but just FYI, you should typically never need to directly access the HTML element to show/hide it when working in an SPA. One of the main benefits of modern SPAs is that they abstract away finicky HTML + JS work. In Angular's case, you'd go into the view and use an `*ngIf` directive to show/hide a UI component

src/app/projects/projlist-filters/projlist-filters.component.ts
```diff
@@ -259,6 +303,18 @@ export class ProjlistFiltersComponent implements OnInit, OnChanges, OnDestroy {
       item._id.toString().indexOf(this.dispIdFilter.toString()) > -1
     );
 
+		// If there are any filters checked, check for matching project type filters
+		if (this.projectTypeFilters.find(ptf => true === ptf.checked)) {
```
Baelx: The `filter` and `map` array methods should be able to make lines 307 - 314 into this:
```suggestion
    // Filters by only those that are checked, then only returns the name for each one instead of the object
		const validatedLocations = this.projectTypeFilters.filter(ptf => ptf.checked).map(ptf => ptf.name);
```

Baelx (changes requested):
Thanks for this! A couple things could could use some tweaking. I'll see if I can use this PR. If not, I'll just create a new one

Baelx (approved):

NatSquared (approved):
LGTM! Thanks both for your work on this

PR #127, opened 2025-01-29
DESENG-752 Remove eventlistener from filter collapse handler

Baelx:
Resolves a breaking bug during build


jareth-whitney (approved):
Yes, we shouldn't need an event to manage that.

PR #128, opened 2025-01-29
DESENG-752 Correct bug with older projects

Baelx:
Older projects without projectTypes were causing the filter to break. Corrected


jareth-whitney (approved):
I'm guessing this is necessary for legacy compatibility.

PR #133, opened 2025-02-27
bugfix/deseng771: Removed zoom/bound reset when modifying filters on project map.

jareth-whitney:
- Removed zoom/bound reset when modifying filters on project map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 1.18.2 Feb 27, 2025
+
+- Disabled bound/zoom reseting after a filter is changed on the project map. [DESENG-771](https://apps.itsm.gov.bc.ca/jira/browse/DESENG-771)
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 1.18.2 Feb 27, 2025
```
Baelx, line 1: I feel like we maybe need to ditch version numbers for LUP. It's getting hazy what's actually getting to production and what's not. 

It may be another year before this app is shelved completely. I think in a future PR we should remove version numbers to lessen confusion (especially if we're going to implement pipelines at some point)

Baelx (approved):
Thanks for the quick fix!

CHANGELOG.md
jareth-whitney: Yes let's discuss this and make a plan.

PR #134, opened 2025-03-12
feature/deseng770: Added BC Basemap layer to all maps

jareth-whitney:
- Added BC Basemap layer to all maps
- Fixed ocean base layer
- Fixed zooming of markers on projects
- Updated Leaflet to version 1.5.0

### DIFF
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -1,8 +1,7 @@
 import { Component, OnInit, AfterViewInit, OnDestroy, ElementRef, ViewChild } from '@angular/core';
 import 'leaflet';
 import 'assets/js/leaflet.shpfile.js';
-import * as Esri from "esri-leaflet";
-import * as EsriVector from "esri-leaflet-vector";
+import { vectorBasemapLayer } from "esri-leaflet-vector";
 
 import { StorageService } from 'app/services/storage.service';
 import { Constants } from 'app/shared/utils/constants';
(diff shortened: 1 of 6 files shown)


src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -14,6 +16,8 @@ import * as _ from 'lodash';
 // need to import leaflet this way to include the shapefile->geojson plugin
 declare let L;
 const encode = encodeURIComponent;
+L.esri = Esri as any;  // Manually attach Esri to L
```
Baelx: Was an effort made to try to use the right type here? I think it would just consist of downloading the right package and finding the right type to use: https://www.npmjs.com/package/@types/esri-leaflet

Baelx (changes requested):
Thanks! Things look good and just wanting to know if/how we can ship code with type safety. Since we don't have unit testing, our type safety is one of the big things that helps reduce bugs or issues with the platform

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: All of the Esri imports have been optimized in the next commit, and all instances of "any" were removed.

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -3,6 +3,7 @@ import { ApplicationRef, ElementRef, SimpleChanges, Injector, ComponentFactoryRe
 import { Subject } from 'rxjs';
 import 'leaflet';
 import 'leaflet.markercluster';
+import { vectorBasemapLayer } from "esri-leaflet-vector";
```
Baelx, line 6: Appreciate the (hopefully) increased type safety. Also liking the smaller code footprint

Baelx (approved):

PR #135, opened 2025-03-13
Feature/deseng770: Added back esri-leaflet-vector dependencies to avoid build fail in Openshift

jareth-whitney:
- Added back esri-leaflet-vector dependencies to avoid build fail in Openshift
- Added required scripts to angular.json


Baelx (approved):
GTG!

PR #137, opened 2025-03-19
Feature/deseng773: Updated project map with snackbar message on search/filter update

jareth-whitney:
- Updated project map with snackbar message on search/filter update

### DIFF
--- a/src/app/projects/projects.component.ts
+++ b/src/app/projects/projects.component.ts
@@ -53,6 +53,7 @@ export class ProjectsComponent implements OnInit, OnDestroy {
   public listApps: Array<Project> = [];
   // private filters: FiltersType = null; // FUTURE
   private ngUnsubscribe: Subject<boolean> = new Subject<boolean>();
+	public filterCount: number = 0;
 
   previousUrl: string;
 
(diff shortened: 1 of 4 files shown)


src/app/projects/projects.component.ts
```diff
@@ -159,9 +160,16 @@ export class ProjectsComponent implements OnInit, OnDestroy {
   /**
    * Event handler called when filters component updates list of matching apps.
    */
-  public updateMatching() {
+  public updateMatching(source: string) {
     // map component gets filtered apps
     this.mapApps = this.filterApps.filter(a => a.isMatches);
+		// Prevent snackbar on initial page load
+		if (2 <= this.filterCount) {
```
Baelx, line 167: If someone loads the page, toggles a filter off, then activates all three filters, will they get feedback? i.e. is there an instance where they could be using three filters after page load?

src/app/projects/projlist-filters/projlist-filters.component.ts
```diff
@@ -39,7 +39,7 @@ export interface FiltersType {
 export class ProjlistFiltersComponent implements OnInit, OnChanges, OnDestroy {
   // NB: this component is bound to the same list of apps as the other components
   @Input() projects: Array<Project> = []; // from projects component
-  @Output() updateMatching = new EventEmitter(); // to projects component
+  @Output() updateMatching = new EventEmitter<string>(); // to projects component
```
Baelx, line 42: Thanks for the added typing!

Baelx (changes requested):
Just wanting to make sure the feedback is given for all searches after page load. Please see my comment! Otherwise, everything looks good

src/app/projects/projects.component.ts
jareth-whitney: Oh yes, this isn't a limit to active filters. It prevents the snackbar from showing up when the page initially loads. Basically, it's saying wait until the third call to this code before showing the snackbar.

Baelx (approved):
Thanks for the explanation of that variable! Good to go

PR #138, opened 2025-03-21
feature/deseng776: Included build and deploy to dev pipeline yaml.

jareth-whitney:
- Included build and deploy to dev pipeline yaml


Baelx (changes requested):
Thanks! Many of my questions from the API pipeline apply here. Let's touch base on those items before proceeding with this and the other PRs.

Baelx (approved):
After discussing this with Steve, we still need to have some kind of notification when Deployments aren't successful. I've been told to approve for now and we may make changes to this work later.

PR #141, opened 2025-04-01
Bugfix/deseng786: Minor code changes to improve sorting and checks on project list page.

jareth-whitney:
- Minor code changes to improve sorting and checks on project list page

### DIFF
--- a/src/app/projects/project-list/project-list.component.ts
+++ b/src/app/projects/project-list/project-list.component.ts
@@ -351,14 +351,14 @@ export class ProjectListComponent implements OnInit, OnDestroy {
     return overlappingDistrictsListString;
   }
 
-	stringifyProjectTypes(projectTypes) {
-		if (!projectTypes || !Array.isArray(projectTypes)) {
-			return '';
-		}
-		const filteredProjectTypes = projectTypes.filter(pt => true === pt.checked);
-		const projectTypeNames = filteredProjectTypes.map(pt => pt.name)?.sort();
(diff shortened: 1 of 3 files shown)


Baelx (approved):
Please just remove the optional chaining before the `sort` call before proceeding!

src/app/projects/project-list/project-list.component.ts
```diff
@@ -356,7 +356,7 @@ export class ProjectListComponent implements OnInit, OnDestroy {
 			return '';
 		}
 		const filteredProjectTypes = projectTypes.filter(pt => true === pt.checked);
-		const projectTypeNames = filteredProjectTypes.map(pt => pt.name);
+		const projectTypeNames = filteredProjectTypes.map(pt => pt.name)?.sort();
```
Baelx: I don't think the optional chaining is necessary. If `map` exists then `sort` will exist. The `map` call will only ever return an array. Also you check for this being an array above in this function.

Just an FYI, the Typescript way to do this would be to set the projectTypes parameter as an array. Then you wouldn't have to worry about what type it'll be.

Baelx (changes requested):
Please see my comment!

Baelx (approved):

src/app/projects/project-list/project-list.component.ts
jareth-whitney: I actually did receive an error related to this line. I believe the issue arrises when the .filter() receives zero results, and then we try to map upon that. If you want me to remove it, I can.

PR #144, opened 2025-04-10
[DESENG-788] Improve project list screen reader accessibility

NatSquared:
https://citz-gdx.atlassian.net/browse/DESENG-788

Fixed issues:
☑️ The column headers are read as headers when using a screen reader
☑️ The column headers announce their sort actions. `${entry.name} column; click to sort`
☑️ The column headers are “clickable” with keyboard using the enter key.
☑️ When using a screenreader, the project listings read the following message aloud: `View more information about the ${projName} project. Link.`;
☑️The project list entry in the table redirects to the named project when the enter key is pressed.
☑️Tested & verified using NVDA screen reader.


Baelx (approved):
You may just need to pull in Jareth's changes. LGTM

PR #145, opened 2025-04-10
bugfix/deseng787: Increased strictness of sorting and pagination in project document tab.

jareth-whitney:
- Increased strictness of sorting and pagination in project document tab.


NatSquared (approved):
LGTM!

PR #146, opened 2025-04-14
Feature/deseng789: Added attachments logic to contact form, updated models and services accordingly

jareth-whitney:
- Added attachments logic to contact form
- Updated models and services accordingly


src/app/project/contact-form/contact-form.component.scss
```diff
@@ -0,0 +1,48 @@
+.successful-send {
```
Baelx, line 1: I've been guilty forgetting this but I think if you use the `git mv` command, you can tell git that a file was moved instead of being deleted and recreated.

Baelx (approved):
Thanks for the doc blocks and typed code. Looks good!

src/app/project/contact-form/contact-form.component.scss
jareth-whitney: I changed the name of the file because it went from CSS to SCSS so I don't think move would work in this instance.

PR #148, opened 2025-04-25
DESENG-769 Add project-level shapefile colour reference to map page

Baelx:


jareth-whitney (approved):
Looks like another colour value fallback, no issues found.

PR #151, opened 2025-05-14
Bugfix/deseng822: Made modifications to comments component HTML and SCSS as per Steve

jareth-whitney:
- Made modifications to comments component HTML and SCSS as per Steve
Ticket: https://citz-gdx.atlassian.net/browse/DESENG-822


NatSquared (approved):
LGTM!

PR #153, opened 2025-05-15
DESENG-821 Fix incorrect shapefile ordering issue

Baelx:


NatSquared (approved):
Looks good! Thanks for fixing up the changelog too

PR #156, opened 2025-05-30
DESENG-825 Add option to hide shapefiles on main map page

Baelx:


NatSquared (approved):
Looks good; aligns with bcgov/landuseplanning-admin#146

PR #159, opened 2025-06-12
bugfix/deseng830: Added additional checking for project list map.

jareth-whitney:
Added additional checking for project list map.


NatSquared (approved):
Looks good! Seems like some logic tweaks and fixes for base cases and edge cases

PR #161, opened 2025-06-18
Bugfix/deseng831: Added additional safeguard for shape file values on individual project page

jareth-whitney:
- Added additional safeguard for shape file values on individual project page
Ticket: https://citz-gdx.atlassian.net/browse/DESENG-831


NatSquared (approved):
LGTM!

PR #163, opened 2025-06-20
bugfix/deseng831: Added shape file support for legacy projects, refactored individual project map code.

jareth-whitney:
- Added shape file support for legacy projects
- Refactored individual project map code
- Changed resolver values to static to guard against external param influence

### DIFF
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -183,8 +183,8 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
       // Draw project-level shape files if they're available.
       shapefilesToDraw = _.orderBy(this.project.shapefiles, ['order'], ['desc']);
     } else if (this.shapefiles.length > 0) {
-      // Otherwise, draw the shape files that were retrieved by the document resolver.
-      shapefilesToDraw = this.shapefiles;
+      // Otherwise, draw the first shape file that was retrieved by the document resolver, if it exists.
+      shapefilesToDraw = [this.shapefiles[0]];
     }
 
 		// Convert documents to shapefiles
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -34,7 +34,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
 	readonly defaultBoundsObject = {southWest: {lat: 48, lng: -139}, northEast: {lat: 61, lng: -114}}; // Converted for other parsing
 	public bounds = {southWest: {lat: null, lng: null}, northEast: {lat: null, lng: null}}; // Bounds object for keeping track of bounds
   private ngbModal: NgbModalRef = null;
-  public shapefiles: Document[][] = [];
+  public shapefiles: Document[] = [];
 	public convertedShapefiles = [];
   public pathAPI: string;
 
@@ -68,6 +68,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
 				})
       }
     });
+
     // The following items are loaded by a file that is only present on cluster builds.
     // Locally, this will be empty and local defaults will be used.
     const remoteApiPath = window.localStorage.getItem('from_public_server--remote_api_base_path');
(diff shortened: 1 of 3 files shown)


src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -177,11 +178,13 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
...
-    let shapefilesToDraw = this.shapefiles[0] || [];
+    let shapefilesToDraw = [];
     if (Array.isArray(this.project.shapefiles) && this.project.shapefiles.length > 0) {
-      shapefilesToDraw = this.project.shapefiles;
-      shapefilesToDraw = _.orderBy(shapefilesToDraw, ['order'], ['desc']);
+      // Draw project-level shape files if they're available.
+      shapefilesToDraw = _.orderBy(this.project.shapefiles, ['order'], ['desc']);
+    } else if (Array.isArray(this.shapefiles) && this.shapefiles.length > 0) {
+      // Otherwise, draw the shape files that were retrieved by the document resolver.
+      shapefilesToDraw = this.shapefiles;
```
Baelx: In the past, there would only ever be one intended shapefile per project. This will show all retrieved shapefiles

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -220,53 +227,62 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
...
 						// Change map bounds based on values in bounds variable
 						if (this.bounds.southWest.lat && this.bounds.southWest.lng && this.bounds.northEast.lat && this.bounds.northEast.lng) {
 							this.map.fitBounds([[this.bounds.southWest.lat, this.bounds.southWest.lng], [this.bounds.northEast.lat, this.bounds.northEast.lng]], {padding: [50, 50]});
-						} else if (!locationAdded && this.project) {
-							this.addMarker();
 						}
 
-						// Add the feature group to the map
-						this.map.addLayer(this.appFG);
+            if (!locationAdded) {
```
Baelx, line 245: How would `locationAdded` ever be false here?

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -220,53 +227,62 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
...
+              // Add a marker if no shape files have been added to the map
+              this.addMarkerAndAdjustBounds();
```
Baelx, line 247: The comment says "if no shape files have been added to the map" but this logic only executes on the last iteration so won't there potentially be a bunch of shapefiles added already?

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -177,11 +178,13 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
...
 
-    // Draw project-level shapefiles if they're available.
-    let shapefilesToDraw = this.shapefiles[0] || [];
+    let shapefilesToDraw = [];
     if (Array.isArray(this.project.shapefiles) && this.project.shapefiles.length > 0) {
-      shapefilesToDraw = this.project.shapefiles;
-      shapefilesToDraw = _.orderBy(shapefilesToDraw, ['order'], ['desc']);
+      // Draw project-level shape files if they're available.
+      shapefilesToDraw = _.orderBy(this.project.shapefiles, ['order'], ['desc']);
+    } else if (Array.isArray(this.shapefiles) && this.shapefiles.length > 0) {
```
Baelx: Was an array not being returned here? That would be unusual. The resolver should at least return an array unless there's some kind of error

Baelx (changes requested):
Thanks for this! I just have a few questions about the logic. Also, I'm curious why this was only happening in Edge and not in other browsers

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: This was the major bug that was preventing legacy projects from working. Although there may have been a single shapefile, it was being fed as an object instead of an array containing an object, which was preventing it from being used. This is the way that it needs to be structured.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: I've revised the logic so it makes more sense now. I will message you with details.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: Yes potentially. The issue was that the existing flags were not doing proper checks after the shape file was added to the feature group (this.appFG), so the checks would always return false. I've now revised that so this is a legitimate fallback if the shape files couldn't be added to the feature group.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: OK I'll take out the first part of the conditional and leave the second in.

Baelx (approved):
Thanks for the updates!

PR #164, opened 2025-06-23
feature/deseng791: Added loading animation to hide map rendering, refined individual project map rendering logic further.

jareth-whitney:
- Added loading animation to hide map rendering
- refined individual project map rendering logic further

Ticket: https://citz-gdx.atlassian.net/browse/DESENG-791

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ### Jun 19, 2025
 
+- Added loading animation to hide map rendering. [DESENG-791](https://citz-gdx.atlassian.net/browse/DESENG-791)
+
+### Jun 19, 2025
+
 - Document the use of image tags in openshift Deployments. [DESENG-841](https://citz-gdx.atlassian.net/browse/DESENG-841)
 - Added support for shape files from legacy projects on individual project page [DESENG-831](https://citz-gdx.atlassian.net/browse/DESENG-831)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-### Jun 19, 2025
+### Jun 23, 2025
 
 - Added loading animation to hide map rendering. [DESENG-791](https://citz-gdx.atlassian.net/browse/DESENG-791)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Jun 23, 2025
 
 - Added loading animation to hide map rendering. [DESENG-791](https://citz-gdx.atlassian.net/browse/DESENG-791)
+  - Added utility function for measuring connection speed
 
 ### Jun 19, 2025
 
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -251,7 +251,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
 
 					// Last iteration only
 					if (analyzedShapefiles === totalShapefiles) {
-            this.runShapeFilesAnalyzedActions(locationAdded);
+            this.shapeFilesLoadedHook(locationAdded);
 					}
 				});
 
@@ -265,7 +265,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
 
           // Last iteration only
 					if (analyzedShapefiles === totalShapefiles) {
-            this.runShapeFilesAnalyzedActions(locationAdded);
+            this.shapeFilesLoadedHook(locationAdded);
 					}
         });
 			});
@@ -280,7 +280,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
    * 
    * @param locationAdded A boolean that indicates whether or not at least one shapefile has been added to the map
    */
-  private runShapeFilesAnalyzedActions = (locationAdded: boolean) => {
+  private shapeFilesLoadedHook = (locationAdded: boolean) => {
     // Change map bounds based on values in bounds property
     if (this.bounds.southWest.lat && this.bounds.southWest.lng && this.bounds.northEast.lat && this.bounds.northEast.lng) {
       this.map.fitBounds([[this.bounds.southWest.lat, this.bounds.southWest.lng], [this.bounds.northEast.lat, this.bounds.northEast.lng]], {padding: [50, 50]});
--- a/src/app/project/project-details-tab/project-details-tab.component.ts
+++ b/src/app/project/project-details-tab/project-details-tab.component.ts
@@ -12,6 +12,7 @@ import { Document } from 'app/models/document';
 import { ProjectShapefile } from 'app/models/project';
 import * as _ from 'lodash';
 import { Utils } from 'app/shared/utils/utils';
+import { animate, style, transition, trigger } from '@angular/animations';
 
 // need to import leaflet this way to include the shapefile->geojson plugin
 declare let L;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ### Jun 19, 2025
```
Baelx: This should be Jun 23

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -170,6 +173,18 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
       }
     }
 
+    // Add "map loaded" listener for removing loading message
+    // Extra timeout is needed to give vector tiles time to render
+    // Base the timeout delay on user connection speed
+    this.map.once('moveend', () => {
```
Baelx, line 179: Won't this event fire every time the user moves or zooms the map? What about using the `load` event?

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -269,6 +275,26 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
 		}
 	}
 
+  /**
+   * The actions that should be run after all shape files are analyzed.
+   * 
+   * @param locationAdded A boolean that indicates whether or not at least one shapefile has been added to the map
+   */
+  private runShapeFilesAnalyzedActions = (locationAdded: boolean) => {
```
Baelx: This would describe what's happening inside the function so you could know at a glance when seeing it get executed:
```suggestion
  private calculateBoundsAndAddMarkers = (locationAdded: boolean) => {
```

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -313,7 +339,7 @@ export class ProjectDetailsTabComponent implements OnInit, AfterViewInit, OnDest
 					? shapefileBounds._northEast.lat 
 					: this.bounds.northEast.lat,
 				lng: shapefileBounds?._northEast.lng <= this.defaultBoundsObject.northEast.lng 
-					&& (!this.bounds.northEast.lng || shapefileBounds?._northEast.lng >= this.bounds.northEast.lat) 
+					&& (!this.bounds.northEast.lng || shapefileBounds?._northEast.lng >= this.bounds.northEast.lng)
```
Baelx, line 342: Were the bounds not being calculated properly in this function? Or was a change needed due to recent work?

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -179,6 +182,18 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
 			}
 		}
 
+    // Add "map loaded" listener for removing loading screen
+    // Must add additional timeout to account for vector tile rendering
+    // Base the timeout delay on user connection speed
+    this.map.once('moveend', () => {
+      this.utils.getConnectionTier().then((tier) => {
+        // Convert the returned connection speed to a millisecond value
+        const delay = { slow: 3500, medium: 3000, fast: 2500, turbo: 2000 }[tier];
```
Baelx, line 191: So it looks like you went with a kind of "last resort" implementation as described in the ticket. Why not use the `map` loop that goes through each shapefile and does a `hasLayer` check? Wouldn't that be at or near the end of all the map rendering events?

src/app/shared/utils/utils.ts
```diff
@@ -49,4 +51,40 @@ export class Utils {
...
+
+  /**
+   * Returns an estimated connection tier based on downlink.
+   * If downlink can't be found, it falls back to small image load time.
+   * 
+   * @returns {Promise<ConnectionTier>}
+   */
+  public async getConnectionTier(): Promise<ConnectionTier> {
+    // Try to retrieve the connection speed directly
+    const conn = (navigator as any).connection || (navigator as any).mozConnection || (navigator as any).webkitConnection;
```
Baelx, line 63: https://www.slingacademy.com/article/working-with-window-navigator-object-in-typescript/ 

This may have some tips for working with types and `navigator`

src/app/shared/utils/utils.ts
```diff
@@ -49,4 +51,40 @@ export class Utils {
...
+      const speed = conn.downlink;
+      if (speed < 1) return 'slow';
+      else if (speed < 10) return 'medium';
+      else if (speed < 100) return 'fast';
+      else return 'turbo';
+    }
+
+    // Image load time fallback
+    return new Promise<ConnectionTier>((resolve) => {
+      const img = new Image();
```
Baelx, line 75: So each time the map moves or zooms, and image is being created and loaded in memory and a setTimeout is fired. I feel this could be a performance concern

src/app/shared/utils/utils.ts
```diff
@@ -49,4 +51,40 @@ export class Utils {
...
+      img.onload = () => {
+        const duration = performance.now() - start;
+        if (duration > 1500) return resolve('slow');
+        else if (duration > 800) return resolve('medium');
+        else if (duration > 400) return resolve('fast');
+        else return resolve('turbo');
+      };
+
+      img.onerror = () => resolve('medium'); // assume average
+      img.src = 'https://www.google.com/images/phd/px.gif?' + Date.now();
```
Baelx, line 85: Curious about URL, I've never heard of this. Where did you get it?

Baelx (changes requested):
Thanks! I appreciate the effort but I have a couple of concerns:

- I think there could be performance concerns when it comes to using that map event in conjunction with spoofing an image and firing a setTimeout
- The Changelog date should be updated to today
- It looks like you went with the "last resort" method. I'm curious if there are other options for us here instead of trying to guess how much time to show the loading screen

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: Not when .once is used. If .once is used then it only fires once. My other call was using .on instead of .once, so I have changed that. As for moveend vs load, moveend triggers up to 500 ms later and is the last loading hook I can grab in the chain of events. That's why it's being used.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: It also adds a layer to a feature group. So yes, I could give it a name like calculateBoundsAndAddMarkersOrAddLayerToFeatureGroup(), or I could title it based on the hook that we're using. I'll simplify it further to something like shapeFilesLoadedHook()

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: It was incorrect but it wasn't of consequence. That's why it was easily missed.

src/app/projects/projlist-map/projlist-map.component.ts
jareth-whitney: No, not really. I went with a hook that fires at the very end of the line (moveevent). It fires once, but then the vector tiles need additional time to render, and there is no hook that we can use to wait for that.

src/app/shared/utils/utils.ts
jareth-whitney: Thanks, I will take a look at this.

src/app/shared/utils/utils.ts
jareth-whitney: No, that's incorrect. Both events are fired with .once now, meaning they're only used once.

src/app/shared/utils/utils.ts
jareth-whitney: You can see an example of it here:
https://stackoverflow.com/questions/5529718/how-can-i-detect-internet-speed-in-javascript

Basically it's an extremely small file, like 40 bytes or something, and it's a 1x1 transparent pixel. It's hosted by Google so it's extremely reliable (up-time), and the download rate that it serves will be consistent.

Finally, keep in mind that this is just a backup if the primary method of speed test doesn't work.

jareth-whitney: Just a small update. I have tested the backup speed testing logic and adjusted it to accomodate observed behaviour in Firefox and Safari. I will do further testing through Firefox and Safari in the dev environment as well. I also observed network traffic in these browsers to make sure that no image fetches were occurring after first load. Everything is behaving as expected.

Baelx (approved):
Approved! Thanks for the discussions

PR #165, opened 2025-06-24
feature/deseng791: Added stricter typing for connection speed utility function

jareth-whitney:
- Added stricter typing for connection speed utility function

Ticket: https://citz-gdx.atlassian.net/browse/DESENG-791


Baelx (approved):
This is great! Thanks for the type safety here. Because we don't run tests, our type safety is basically our only automated guard rails so it makes a difference to have type certainty.

PR #169, opened 2025-07-03
DESENG-866 Restore image picker to survey WYSIWYG editors

Baelx:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### Jul 3, 2025
+- Restore use of the image picker in survey WYSIWYG editors [DESENG-866](https://citz-gdx.atlassian.net/browse/DESENG-866)
+- Remove some unused code
```
jareth-whitney: Which code did you remove? Elaborating on this comment will help us track down which code was removed later.

src/app/project/comments/add-survey-response/add-survey-response.component.scss
```diff
@@ -181,6 +180,18 @@ app-file-upload {
...
 }
 
+.question, .info {
+    // Access elements that have been inserted via innerHTML
+    ::ng-deep {
+      img {
+          display: block;
+          margin: 0 auto;
+          max-width: 100%;
+          min-width: 50px;
```
jareth-whitney, line 190: Does this min-width value look ok on mobile?

src/app/project/comments/add-survey-response/add-survey-response.component.scss
```diff
@@ -181,6 +180,18 @@ app-file-upload {
   display: flex;
 }
 
+.question, .info {
```
jareth-whitney, line 183: When using ng deep, it is often beneficial to be as specific as possible with your CSS, example: .info.control-label instead of .info.

jareth-whitney (approved):
This one looks pretty simple, just some CSS/HTML changes. Please review the following before you merge:

- Elaborate on which code was removed in changelog
- Specificity of CSS selector
- image min-width value on mobile

src/app/project/comments/add-survey-response/add-survey-response.component.scss
Baelx: I would understand if you wanted to use specific, long class names for content entered into the WYSIWYG editor by the user. But the only way this could collide, that I can see, would be if a developer specifically used the .info class within the associated HTML file. Is there something I'm missing here?

src/app/project/comments/add-survey-response/add-survey-response.component.scss
Baelx: You're thinking it might be too small for mobile? I can double check

CHANGELOG.md
Baelx: I feel like if we detailed every place we removed commented-out code, it would get a bit too busy in the Changelog. I can definitely update this to "removed some commented-out code" so there's no impression that chunks of non-commented code were removed

PR #170, opened 2025-07-10
bugfix/deseng839: Added marker/centroid validation to project list map and individual project map.

jareth-whitney:
Added marker/centroid validation to project list map and individual project map.

Ticket: https://citz-gdx.atlassian.net/browse/DESENG-867

### DIFF
--- a/src/app/projects/projlist-map/projlist-map.component.ts
+++ b/src/app/projects/projlist-map/projlist-map.component.ts
@@ -354,7 +354,7 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
         })
       } else {
         // If no shapefile is found for a project, display a pin of its coordinates instead.
-        if (2 === proj.centroid.length) {
+        if (this.markerMeetsConditions(proj)) {
           const title = `${proj.name}\n`
           + `${proj.overlappingRegionalDistricts}\n`;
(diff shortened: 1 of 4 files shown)


src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -372,6 +372,22 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
     this.setVisibleDebounced();
   }
 
+  /**
```
Baelx: Thanks for the additional safety checks to prevent a crash! Could you put this under `shared/utils/utils.ts`? It's identical to the other one right?

Baelx (changes requested):
Thanks for the frontend safety checks. If those two functions are identical, could you please consolidate them into a utils file like `shared/utils/utils.ts`? This can save us a some bytes for later when we look at reducing bundle size

src/app/projects/projlist-map/projlist-map.component.ts
jareth-whitney: I was originally going to do that. However, one component has one argument, while the other has none. In one case, you are using a class property, and in the other case you are not. So I have to use this.project as an argument if I unify the methods, which felt wrong. That's ok though, I'll just feed this.project in as an argument and make the function shareable.

Baelx (approved):
Thanks!

PR #171, opened 2025-07-22
feature/deseng865: Improved accessibility, design, and compliance of header/footer and other site content

jareth-whitney:
- Updated the header/footer with improved design, accessibility, and inclusivity. [DESENG-865](https://citz-gdx.atlassian.net/browse/DESENG-865)
  - Added a skip link to the header
  - Added current page menu highlighting
  - Made the header/footer mobile compatible
- Made all site content mobile compatible
- Adjusted margins of pages that were not aligned with header/footer
- Changed some page content to improve accessibility (contrast, inverse text)
- Added external link icons beside all external links
- Added additional hover events for increased site interactivity
- Removed all instances of colonial blue
- Redesigned project contact info / contact button / subscribe button area to separate it from the footer
- Restructured SCSS for increased specificity and legibility
  - Changed named colours like 'black' or 'white' to hex codes like '#000' or '#fff'

Ticket: https://citz-gdx.atlassian.net/browse/DESENG-865

** Apologies for PR size, as per our contribution guide. Many of the files just have one line of changes, related to CSS colour value replacement **


src/app/contact/contact.component.html
```diff
@@ -42,9 +42,18 @@ <h2>Contact the Program Teams</h2>
         </span>
         <h2>Learn more about provincial land and water planning</h2>
         <p>Find more information about</p>
-        <a class="paragraph-link" href="https://www2.gov.bc.ca/gov/content/industry/crown-land-water/land-use-planning">Land Use Planning</a>
-        <a class="paragraph-link" href="https://www2.gov.bc.ca/gov/content/industry/forestry/managing-our-forest-resources/forest-landscape-plans">Forest Landscape Planning</a>
-        <a class="paragraph-link" href="https://www2.gov.bc.ca/gov/content/environment/air-land-water/water/water-planning-strategies">Water Planning</a>
+        <a class="paragraph-link" href="https://www2.gov.bc.ca/gov/content/industry/crown-land-water/land-use-planning">
+          Land Use Planning
+          <img class="external-link" src="/assets/images/external-link.png" alt="" />
```
Baelx: This icon is typically used to tell users that the link is going to open in a new tab or window. The icon is meant to be a more minimal replacement for this accessibility technique: https://www.w3.org/TR/WCAG20-TECHS/G201.html

Also, it'll only be perceivable to people without visual impairments. It's a nice thought to have a visual indicator of where the link goes. Maybe you could put (gov.bc.ca) in the anchor tag text? But accessibility-wise, it'd be fine to have no icon here and open the link in the same tab.

src/app/contact/contact.component.scss
```diff
@@ -1,86 +1,100 @@
...
+        flex-direction: column;
+      }
 
-  a:hover {
-    text-decoration: none;
-  }
+      h2 {
+        font-size: 1rem;
+        font-weight: 700;
+        // text-transform: uppercase;
```
Baelx:
```suggestion
```

src/app/contact/contact.component.scss
```diff
@@ -99,3 +113,9 @@
     text-decoration: none;
   }
 }
+
+.external-link {
+  padding-bottom: 10px;
+  // height: 10px;
```
Baelx: Just be sure to remove this commented-out code

src/app/engagement/engagement.component.html
```diff
@@ -27,7 +27,10 @@ <h1 tabindex="-1">Water Planning</h1>
         </p>
         <p>
           The
-            <a style="font-style: italic "href="https://www.bclaws.gov.bc.ca/civix/document/id/complete/statreg/14015">Water Sustainability Act</a>
+            <a style="font-style: italic "href="https://www.bclaws.gov.bc.ca/civix/document/id/complete/statreg/14015">
```
Baelx, line 30: Here's a recommended technique for giving users more information about where they're going when they click a link: https://www.w3.org/WAI/WCAG21/Techniques/html/H33

It involves using the `title` attribute of the a tag

src/app/footer/footer.component.html
```diff
@@ -11,3 +11,93 @@
...
+            </a>
+          </li>
+        </ul>
+      </section>
+    </div>
+
+  </div>
+  <div class="footer-bottom-row">
+
+    <section id="copyright">
```
Baelx: Is there a copyright for this site? For what, the name?

src/app/footer/footer.component.scss
```diff
@@ -3,101 +3,135 @@
...
+      @media (max-width: 720px) {
+        width: 100%;
+      }
+
+      section {
+        width: 100%;
+        line-height: 2;
+
+        a, a:visited {
+          // color: #fed262 !important;
```
Baelx: Some commented-out code to be removed

src/app/footer/footer.component.scss
```diff
@@ -3,101 +3,135 @@
...
+}
 
-  p {
-    font-size: 0.875rem;
-    line-height: 1.5;
-  }
+p.footer-paragraph {
+  padding: 0;
+  margin: 0 0 1rem;
+  line-height: 1.75 !important;
```
Baelx: Is this `!important` necessary? Didn't you just add this paragraph?

src/app/footer/footer.component.scss
```diff
@@ -137,7 +172,15 @@
   }
 
   ul {
+    width: 1140px;
+    max-width: 100%;
+    margin: 0 auto !important;
```
Baelx: I'd recommend using a more granular selector to override other margin styles instead of `!important`

src/app/header/header.component.html
```diff
@@ -1,66 +1,94 @@
...
-            <button type="button" class="close" data-dismiss="alert" aria-label="Close">
-              <span aria-hidden="true">&times;</span>
-            </button>
+
+          </li>
+        </ul>
+      </nav>
+      
+      <div class="alert alert-danger alert-dismissible fade show" id="browser-alert" role="alert">
+        <i class="material-icons md-24">error</i><strong>  Attention: </strong>This website is not supported by Internet Explorer and Microsoft Edge, please use Google Chrome or Firefox.
```
Baelx: For sure we don't need to support Internet Explorer, but we need to make as much effort as the government to support all modern, commonly-used browsers. Are there issues with Microsoft Edge that we need to address?

Otherwise, if you're just rearranging the code around this message, I'd say please just remove it. We can just take it as a given that Internet Explorer isn't used and that Microsoft Edge is a mature product

src/app/header/header.component.scss
```diff
@@ -1,13 +1,134 @@
...
+
+      // Navbar Toggle Button
+      .navbar-toggler {
+        padding: 0.5rem 0.4rem;
+        margin-left: auto;
+        border: 2px solid grey !important;
+        color: $header-font-color !important;
+        font-size: medium;
+        font-weight: 600;
+        outline: none; // override user agent stylesheet default
```
Baelx: Comments are generally appreciated, but I'm not sure why very select declarations get comments and other don't. Much of the CSS we write is to override the user agent stylesheet. Why is this one especially noteworthy to developers?

src/app/header/header.component.scss
```diff
@@ -1,13 +1,134 @@
...
+
+      @media (max-width: 1200px) {
+        width: 100%;
+      }
+
+      // Navbar Toggle Button
+      .navbar-toggler {
+        padding: 0.5rem 0.4rem;
+        margin-left: auto;
+        border: 2px solid grey !important;
```
Baelx: Which CSS declaration is conflicting with this style? Why was !important necessary here? Can you make this an ID instead of a class for greater specificity?

src/app/header/header.component.scss
```diff
@@ -110,136 +168,162 @@
...
 
-      .icon {
-        margin-top: 0;
+    .nav-item {
+      display: flex;
+      flex-direction: row;
+
+      .nav-link {
+        position: relative;
+        z-index: 1001; // +1 to keep above dropdown menus
```
Baelx, line 178: This makes sense to me as a useful CSS comment to have for developers :)

src/app/header/header.component.scss
```diff
@@ -110,136 +168,162 @@
...
+          }
+        }
+
+        &:hover,
+        &:focus {
+
+          >a,
+          >a:hover,
+          >a:focus {
+            // background-color: $header-bg;
```
Baelx: Commented-out code

src/app/header/header.component.ts
```diff
@@ -1,25 +1,33 @@
...
   constructor(
     public router: Router
   ) { }
   // tslint:disable-next-line:use-life-cycle-interface
   ngOnInit() {
-  let isIEOrEdge = /msie\s|trident\/|edge\//i.test(window.navigator.userAgent);
-  const browser_alert = document.getElementById('browser-alert');
-  if ( isIEOrEdge) {
-    browser_alert.classList.add('showForIEorEdge');
+    let isIEOrEdge = /msie\s|trident\/|edge\//i.test(window.navigator.userAgent);
```
Baelx: This variable may no longer be necessary. We should be supporting Edge now and it's a given that Internet Explorer users are going to have a bad time :P

src/app/header/header.component.ts
```diff
@@ -30,6 +38,14 @@ export class HeaderComponent {
...
 
+  public scrollTo(targetId: string): void {
+    const element = document.getElementById(targetId);
+    if (element) {
+      this.router.navigate([], { fragment: targetId });
+      element.scrollIntoView({ behavior: 'smooth' });
+    }
+  }
+
   // const learnMore = document.getElementById('aboutMMTI');
```
Baelx: Would be appreciated if you could remove this commented-out code while you're here. But not a blocker to the PR

src/app/home/home.component.html
```diff
@@ -102,27 +102,29 @@ <h2>About this Site</h2>
            and the Biodiversity and Ecosystem Health Framework may be added to this site over time.
         </p>
         <h2 class="learn-more-heading">Learn More</h2>
-        <div class="row feature-blocks">
-          <div class="col-md-4">
-            <div class="feature-block">
+        <div class="feature-blocks">
+          <div class="feature-block-outer">
+            <!-- <img src="/assets/images/lup_revelstoke.jpg" class="feature-block-image land" aria-hidden="true" /> -->
```
Baelx: Commented-out code in this file should be removed.

src/app/home/home.component.scss
```diff
@@ -36,37 +53,116 @@ main {
   p {
     line-height: 1.75;
     font-size: 1.2rem;
-    /*max-width: 780px;*/
```
Baelx, line 39: Thanks for removing this commented-out code!

src/app/project/project.component.html
```diff
@@ -89,27 +89,52 @@ <h3>{{project.commentPeriodForBanner.dateStarted | date: 'MMMM d'}} -
...
-                  href="mailto:{{project?.projectLead?.email}}">{{project?.projectLead?.email || "-"}}</a></p>
-            </ng-template>
-          </div>
-          <div class="inner-col">
-            <button class="btn btn-small inverted" (click)="addEmail()" type="button">Sign up for project updates</button>
+
+        <div class="people-item">
+
+          <button *ngIf="project?.contactFormEnabled" class="btn btn-small contact-anonymous" (click)="launchContactForm()" type="button">
+            <i _ngcontent-c11="" class="material-icons-outlined button-icon" role="presentation" wiid="C937">email</i>
```
Baelx: Often, icons will attempt to be parsed by a screenreader. I'm not sure how material icons handles this, but you may need to add:

```suggestion
            <i _ngcontent-c11="" aria-hidden="true" class="material-icons-outlined button-icon" role="presentation" wiid="C937">email</i>
```

src/app/projects/projlist-filters/projlist-filters.component.html
```diff
@@ -2,23 +2,40 @@
...
+          </button>
         </div>
-        <button type="button" class="btn btn-primary" title="Apply Filters" id="applyButton" (click)="applyAllFilters();"
-        [disabled]="loading"><i class="material-icons" aria-hidden="true">search</i>Search</button>
-      </fieldset>
 
-    </div>
-    <div class="instructions map-instructions">You can also <a class="paragraph-link" routerLink="/projects-list">go to list view of projects.</a></div>
+        <div *ngIf="searchOpen">
+          <p class="additional-text">Additional land and water planning projects will be added to the map as they are initiated. You can also
```
Baelx, line 16: In the interest of accessibility, could we please make the text about the "list view page" more perceivable to screenreaders? Maybe an sr-only label that says something like: "People using assistive devices are advised to search on the list view page," and then provide the link.

src/app/projects/projlist-filters/projlist-filters.component.scss
```diff
@@ -79,22 +99,22 @@ $app-filters-font-lg: 0.9375rem;
     .btn {
       border-radius: 0 4px 4px 0;
       font-size: 1rem;
-      position: absolute;
-      right: 22px;
-      bottom: 19px;
+      // position: absolute;
```
Baelx: Commented-out code

src/app/projects/projlist-filters/projlist-filters.component.scss
```diff
@@ -317,29 +337,41 @@ $app-filters-font-lg: 0.9375rem;
...
+  }
 }
 
-.project-type-filter-title-row {
+.project-type-filter-title-row, .project-search-title-row {
 	display: flex;
 	flex-direction: row;
 	width: 100%;
 	justify-content: space-between;
+  margin-bottom: 1.25rem;
```
Baelx, line 350: What happened to the spacing here?

src/app/share-buttons/share-buttons.component.scss
```diff
@@ -5,6 +5,11 @@
     top: 200px;
     padding: 0px;
 
+    // Avoid blocking important content on mobile
```
Baelx, line 8: Does this hide the social share buttons on mobile? Is the PO okay with that?

Baelx (changes requested):
Thanks for your work here, Jareth! The "skip to content" button, improved mobile experience, and other accessibility improvements are going to be great! A few things to look at!

- Please be sure to remove any commented-out code in all files
- Remove the message about supported browsers
- Double-check the copyright message: does this site have any kind of copyright?
- Any icons that are part of links should have 'aria-hidden="true"' unless they're supposed to be parsed by screenreaders (or Material UI is able to hide those)
- Please confirm that Steve is okay with hiding the social share buttons on mobile and tablet devices
- I really appreciate your consideration for trying to let users know that a URL is off-site. Unfortunately, showing that particular icon is not the way to do it and will make people think the link is opening in a new tab. WCAG recommends using the `title` attribute for giving people advance notice of where they're going. Please see the linked resource article
- There's a prolific use of the `!important` keyword. This is really meant as a last resort and can [make maintenance of the app trickier](https://stackoverflow.com/a/8360237). I count 22 novel uses. Could you please try to get this down to less than 5? I'd recommend using increased selector specificity. Also check in the inspector panel to see which styles are overriding you and why

src/app/contact/contact.component.html
jareth-whitney: I'm not sure what you mean, but it's called an external link icon, and its sole purpose is to tell the user that the link is on an external website. I will include some references.  This was something that was integrated automatically with the WP block theme previously.

https://designsystem.digital.gov/components/link/
https://ux.stackexchange.com/questions/4636/external-links-whether-how-to-distinguishing-them-from-internal-links-and-to

My original instinct was to include alt tags for the images that would be read as part of the link. However, I think auditory clutter should be a big consideration as well. For example, in the footer or on the FAQ page, it would be saying "external link" constantly. If you think it's still best to read the text, I can add it back to the alt tag.

Also, I'm not opposed to making every external link open in a new window, if that's something we want to pursue.

src/app/footer/footer.component.html
jareth-whitney: I ran the footer by Steve and we revised it based on the client's needs, but I'll run this by him just in case.

src/app/project/project.component.html
jareth-whitney: I'll read the icons with a screen reader and see what I get

src/app/projects/projlist-filters/projlist-filters.component.html
jareth-whitney: Sure, I can add a message about that.

src/app/projects/projlist-filters/projlist-filters.component.scss
jareth-whitney: I'm not sure, it looks fine in my editor. I'll just use regular spaces to make sure it works.

src/app/share-buttons/share-buttons.component.scss
jareth-whitney: I gave Steve a full demo, including the full website in mobile, but I am not sure if this was discussed specifically. The way it was setup before, it was actively blocking all of the content in front of it. So if we re-integrated it, it would need to at least stop doing that...

src/app/header/header.component.scss
jareth-whitney: It's existing code, probably nested differently, but I can remove it.

src/app/header/header.component.html
jareth-whitney: This is legacy code that I left in. You're right though, there's no need to support for IE anymore and Edge is Chromium-based, so this code should be irrelevant.

src/app/header/header.component.scss
jareth-whitney: I'll take a look at specificity and see how many !important flags I can remove.

src/app/contact/contact.component.html
Baelx: I don't think name alone is enough to guide our accessibility practices. If you search for "new tab icon" you get many similar results. In practice, this icon has been used interchangeably for new tabs/window, something that commenters in your second link mention. WCAG doesn't have any requirement for identifying external links that I can find. Given this, some organizations, like the UK gov, have removed the icon entirely:

https://design-system.service.gov.uk/styles/links/#external-links

Given the ambiguity of the icon, and possible resulting confusion, my feeling is that we shouldn't be using it.

src/app/share-buttons/share-buttons.component.scss
Baelx: I talked to him about it this morning and he said it's fine to leave out but we'll need to create a ticket to replace it. He said he'll be doing that shortly

src/app/header/header.component.html
Baelx: Great, yeah let's just delete it while we're here since it's not true anymore. Cheers!

src/app/header/header.component.scss
Baelx: Oh okay no biggie either way then

src/app/contact/contact.component.html
jareth-whitney: I did some research of the BC Gov website and the BC Design System and didn't see them there, so let's use that as a decider in this case. I'll remove them based on that.

src/app/footer/footer.component.html
jareth-whitney: Steve has decided that we will not use a copyright notice. A copyright notice can be posted on any original creative work, and can either be registered officially or not. There is no application process required. As of now, the site will not be protected from reuse.

src/app/footer/footer.component.scss
jareth-whitney: It was necessary, in order to override class, but I removed the original class and removed the !important

src/app/project/project.component.html
jareth-whitney: I ended up using aria-hidden="true"

src/app/footer/footer.component.scss
jareth-whitney: I didn't end up needing this CSS at all, since Steve requested the removal of the paragraph in the footer. So I actually just ended up removing this code.

Baelx (approved):
Thanks for the updates! You should be good to go

PR #172, opened 2025-07-24
DESENG-874 Add marker when shapefile is hidden on main map

Baelx:


src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -350,20 +350,14 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
...
-          const title = `${proj.name}\n`
-          + `${proj.overlappingRegionalDistricts}\n`;
-          const marker = L.marker(L.latLng(proj.centroid[1], proj.centroid[0]), { keyboard: true, title: title })
-          .setIcon(markerIconYellow)
-          .on('click', L.Util.bind(this.onMarkerClick, this, proj));
-          marker.projectId = proj._id;
-          this.markerList.push(marker); // save to list
-          this.markerClusterGroup.addLayer(marker); // save to marker clusters group
-        }
+        this.addMarkerToMap(proj);
```
jareth-whitney, line 392: That should tidy things up a bit

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -383,6 +377,26 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
...
 
+  /**
+   * Checks that the marker coordinates are valid and within bounds,
+   * then sets the pop-up message content, creates the marker icon,
+   * adds a click handler for the pop-up, and adds the new layer to the
+   * list of map layers.
+   * 
+   * @param proj The project to add the marker for.
+   */
+  private addMarkerToMap(proj: Project): void {
```
jareth-whitney, line 419: Love the usage of that utility function here.

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -383,6 +377,26 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
...
+    if (this.utils.markerMeetsConditions(proj)) {
+      const title = `${proj.name}\n` + `${proj.overlappingRegionalDistricts}\n`;
```
jareth-whitney: Curious about the usage of \n here. Will this work under all conditions?

jareth-whitney (approved):
Good fallback for the new hidden shape file feature. My only comment is to perhaps review usage of "\n" but I would not say it's blocking a merge.

src/app/projects/projlist-map/projlist-map.component.ts
Baelx: Should be! Template literals support newline characters: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#multi-line_strings

This old logic was moved into its own function and I believe it's been working fine up until now.

PR #174, opened 2025-08-07
bugfix/deseng872: Optimized site loading, enhanced site loading screen, other misc fixes

jareth-whitney:
- Optimized site load times and added a loading/welcome message for limited-speed users. [DESENG-872](https://citz-gdx.atlassian.net/browse/DESENG-872)
  - Measured load times with different speeds and in different environments and added results to spreadsheet
  - Used globs and custom script loading for map scripts instead of app-wide script loading
  - Lazy loaded some components for further optimization, including individual project, project list, and project map
  - Replaced moment.js with dayjs for date handling, which is similar but much less resource-intensive
  - Added post-optimization local load times to spreadsheet for comparison
  - Added a friendly loading/welcome message for people with slow connections
- Renamed misspelled component (activites > activities)
- Increased efficiency of font loading for outlined material icons
- Matched versions for all Leaflet dependencies
- Removed legacy icon fonts for IE and IOS 5


CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+### Aug 6, 2025
+- Optimized site load times and added a loading/welcome message for limited-speed users. [DESENG-872](https://citz-gdx.atlassian.net/browse/DESENG-872)
+  - Measured load times with different speeds and in different environments and added results to spreadsheet
```
Baelx: I appreciate the detail but my feeling is that the Changelog should be a bit more terse than the ticket when it comes to describing the work. Keeping things briefer here will make it more navigable if we need to reference it later. If we're adding 12 lines for each ticket then we're going to quickly end up with an unwieldy Changelog file.

I'm guessing you didn't include a spreadsheet in the code? I think it's great to capture detail in the ticket (and spreadsheets) and then just reference the ticket in the Changelog

angular.json
```diff
@@ -18,34 +18,72 @@
             "polyfills": "src/polyfills.ts",
             "assets": [
               "src/assets",
+              {
```
Baelx, line 21: My understanding of "assets" is that it's meant to contain images, fonts, etc. Why did you opt for scripts and node_modules to be loaded like static assets?

angular.json
```diff
@@ -18,34 +18,72 @@
...
+                "input": "node_modules/leaflet/dist",
+                "output": "leaflet"
+              },
+              {
+                "glob": "leaflet.css",
+                "input": "node_modules/leaflet/dist",
+                "output": "leaflet"
+              },
+              {
+                "glob": "**/*",
```
Baelx: I'm pretty sure this selector would include all CSS files, meaning that you shouldn't need the following selector. Also, there's a text file included in this directory which we shouldn't be including as an asset so I think something like `*.{js,css}` would be fine? We shouldn't need the recursive directory selector either

src/app/app.module.ts
```diff
@@ -52,7 +50,7 @@ import { NewsMultifieldFilterPipe } from 'app/shared/pipes/news-multifield-filte
 import { LegislationComponent } from 'app/legislation/legislation.component';
 import { ProcessComponent } from 'app/process/process.component';
 import { ComplianceOversightComponent } from 'app/compliance-oversight/compliance-oversight.component';
-import { ActivitiesListTableRowsComponent } from './project/project-activites/activities-list-table-rows/activities-list-table-rows.component';
+import { ActivitiesListTableRowsComponent } from './project/project-activities/activities-list-table-rows/activities-list-table-rows.component';
```
Baelx, line 53: Nice catch!

src/app/leaflet/leaflet-augmentations.ts
```diff
@@ -0,0 +1,38 @@
+// Extend leaflet types here
```
Baelx: I think "extending" has a pretty specific meaning in Typescript and refers to things like line 6. I think it'd be more accurate to say you're just declaring module types since Typescript support is incomplete

src/app/models/commentperiod.ts
```diff
@@ -115,20 +117,19 @@ export class CommentPeriod {
...
+      const now = dayjs();
+      const dateStarted = dayjs(obj.dateStarted);
+      const dateCompleted = dayjs(obj.dateCompleted);
+      const sevenDaysBeforeStart = dateStarted.subtract(7, 'day');
 
-      if (moment(now).isBetween(dateStarted, dateCompleted)) {
+      if (now.isBetween(dateStarted, dateCompleted)) {
         this.commentPeriodStatus = 'Open';
-        let days = dateCompleted.diff(moment(now), 'days');
+        const days = dateCompleted.diff(now, 'day');
```
Baelx, line 127: I appreciate you tightening up variable bindings where you can!

src/app/project/project-activities/project-activities.component.html
```diff
@@ -30,13 +30,13 @@
...
     </div>
 
-    <app-table-template *ngIf="!loading && tableParams.totalListItems != 0"
-                        (checkboxChange)='checkChange($event)'
-                        [columns]="tableColumns"
-                        [data]="tableData"
-                        (onColumnSort)='setColumnSort($event)'
-                        (onSelectedRow)='updateSelectedRow($event)'
-                        (onPageNumUpdate)='getPaginatedDocs($event)'>
+    <app-table-template *ngIf="!loading && recentActivities.length > 0"
```
Baelx, line 33: Why was this change necessary? Was `totalListItems` not getting updated?

src/app/project/project-activities/project-activities.component.ts
```diff
@@ -43,21 +46,17 @@ export class ProjectActivitesComponent implements OnInit, OnDestroy {
     private _changeDetectionRef: ChangeDetectorRef) { }
 
   ngOnInit() {
+    // Assign values from inputs
+    this.currentProject = this.project || null;
```
Baelx: I feel like it would make more sense to rename the few instances of `currentProject` to be `project`. Is there a reason we need to keep the `currentProject` property separate? Also if we're expecting it to have the `id` property later in the code, then we should use optional chaining for safety

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -25,69 +22,120 @@ declare let L;
...
-    this.commentPeriod = this.project.commentPeriodForBanner;
-    this.route.data.subscribe((res: any) => {
-      if (Array.isArray(this.project?.shapefiles) && this.project?.shapefiles.length > 0) {
-        return;
+
+    const remoteApiPath = window.localStorage.getItem('from_public_server--remote_api_base_path');
+    this.pathAPI = _.isEmpty(remoteApiPath) ? 'http://localhost:3000/api' : remoteApiPath;
+
+    // Try to get project from parent resolver
+    const parentWithProject = this.route.pathFromRoot.find(r => r.snapshot.data?.projectAndBanner);
```
Baelx: If you're just getting data from the parent resolver, I think it would be clearer to write:

```suggestion
    const projectFromParentResolver = this.route.pathFromRoot.find(r => r.snapshot.data?.projectAndBanner);
```

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -25,69 +22,120 @@ declare let L;
...
+    }
 
-      if (Array.isArray(res?.documents)) {
-				res.documents.forEach(document => {
-					if (document?.data?.meta?.length > 0) {
-						this.shapefiles.push(...document.data.searchResults);
-					}
-				})
+    // Load activities from current route
+    this.route.data.subscribe((res: any) => {
```
Baelx: You should be able to find the exact type being returned in the code elsewhere

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -25,69 +22,120 @@ declare let L;
...
+        this.activities = res.activities;
       }
     });
+  }
 
-    // The following items are loaded by a file that is only present on cluster builds.
-    // Locally, this will be empty and local defaults will be used.
-    const remoteApiPath = window.localStorage.getItem('from_public_server--remote_api_base_path');
-    this.pathAPI = (_.isEmpty(remoteApiPath)) ? 'http://localhost:3000/api' : remoteApiPath;
+  private processShapefilesFromRoute(res: any): void {
```
Baelx: Please add a doc block for brand new functions/methods in LUP! Also, you should be able to track down the parameter type depending on how this method is meant to be used

src/app/project/project-details-tab/project-details-tab.component.ts
```diff
@@ -25,69 +22,120 @@ declare let L;
...
+    ]);
+    // Load subsequent scripts
+    await this.scriptLoader.loadScripts([
+      '/esri-leaflet/esri-leaflet.js',
+      '/esri-leaflet-vector/esri-leaflet-vector.js',
+      '/shpjs/shp.min.js',
+      '/leaflet-shpfile/leaflet.shpfile.js'
+    ]);
+
+    this.L = (window as any).L;
```
Baelx, line 145: I know a lot of this came from old patterns in the app. I'm just trying to understand the current behaviour: so the scripts are loaded into the window and then "L" refers to any leaflet or map-related libraries? Why do we need to make a pointer called `L_` ?

src/app/app-routing.module.ts
```diff
@@ -36,15 +34,14 @@ const routes: Routes = [
...
     path: 'projects',
-    component: ProjectsComponent,
-    data: {
-      title: 'Map View of Projects',
-      focush1: true
-    }
+    loadChildren: () => import('./projects/projects.module').then(m => m.ProjectsModule),
+    data: { 
+      title: 'Map View of Projects', 
+      focush1: true }
```
Baelx:
```suggestion
      focush1: true
    }
```

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -266,10 +288,10 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
     this.updateVisible.emit();
   }
 
-  private fitBounds(bounds: L.LatLngBounds = null) {
-    const fitBoundsOptions: L.FitBoundsOptions = {
+  private fitBounds(bounds: LatLngBounds = this.defaultBounds) {
+    const fitBoundsOptions: FitBoundsOptions = {
       // use top padding to adjust for filters header (which is always visible)
-      // paddingTopLeft: L.point(0, this.appfilters.clientHeight),
+      // paddingTopLeft: this.L.point(0, this.appfilters.clientHeight),
```
Baelx: 🧐

src/app/services/api.ts
```diff
@@ -147,16 +147,22 @@ export class ApiService {
     if (populate !== null) { queryString += `&populate=${populate}`; }
     if (Object.keys(queryModifier).length !== 0) {
       Object.keys(queryModifier).forEach(key => {
-        queryModifier[key].split(',').map(item => {
+        const val = queryModifier[key];
```
Baelx, line 150: Just curious which issue this is solving

src/app/services/scriptLoader.service.ts
```diff
@@ -0,0 +1,94 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({ providedIn: 'root' })
+export class ScriptLoaderService {
+  private loadedScripts: { [url: string]: boolean } = {};
+  private loadedStyles: { [url: string]: boolean } = {};
+
+  /**
+   * Load a single script
```
Baelx: Could you provide a bit more detail in the doc block here? Why would a developer opt to use this instead of the traditional means of loading a script or dependency? Same goes for the function below.

In general, the doc block should give us more information than the title of the function

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -78,67 +50,112 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
     private injector: Injector,
     private resolver: ComponentFactoryResolver,
     private utils: Utils,
+    private scriptLoader: ScriptLoaderService
   ) { }
 
   // create map after view (which contains map id) is initialized
-  ngAfterViewInit() {
-    const self = this; // for closure function below
+  async ngAfterViewInit() {
```
Baelx, line 59: Okay so it looks like you created some utility functions for loading in map-related dependencies. Would creating a separate module and lazy loading it not have been the same thing?

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -78,67 +50,112 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
...
+      '/leaflet-markercluster/MarkerCluster.css',
+      '/leaflet-markercluster/MarkerCluster.Default.css'
+    ]);
+    // Load prerequisite scripts
+    await this.scriptLoader.loadScripts([
+      '/maplibre-gl/maplibre-gl.js',
+      '/leaflet/leaflet.js',
+    ]);
+    // Load subsequent scripts
+    await this.scriptLoader.loadScripts([
```
Baelx: I'm not seeing any error handling here. I feel like we should gracefully notify or move the user somewhere if one of these scripts has an `onerror` event

src/app/services/scriptLoader.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /**
+   * Load multiple scripts
+   * 
+   * @param paths The locations of the scripts that you would like to load
+   * @returns Promise<void>
+   */
+  loadScripts(paths: string[]): Promise<void[]> {
+    return Promise.all(paths.map(path => this.loadScript(path)));
```
Baelx, line 85: Could use a `catch` handler maybe. I think if one fails, nothing will get reported

Baelx (changes requested):
Thanks for the improvements to load times! I'm looking forward to seeing the results. Could you have a look at a few things please:

- The error handling for the load scripts/styles functions is a bit incomplete. I'm not sure what currently happens from the perspective of the user but if something breaks we should try to redirect them back to the homepage or some other user-friendly response. Could you please update the doc blocks for those functions to be more descriptive, also?
- Remove some commented-out code related to points on the map
- Add a doc block for every new function/method
- Remove any added instances of `any` or provide a comment on why it's not feasible
- Have a look at the project-activities component to see if we can cut down on class properties
- Review the commenting in leaflet-augmentations.ts for clarity
- The glob pattern matching in angular.json may not be as efficient as it could be

Thanks!

angular.json
jareth-whitney: The scripts are extracted from node modules and added to the dist folder, so they are available in the compiled version as separate files. If you bundle them with the app as scripts, they are included in the main application bundle and hence take up more space, which leads to slower load times.

angular.json
jareth-whitney: I think you're right, I will try modifying it and compare.

src/app/project/project-activities/project-activities.component.html
jareth-whitney: You have a "no activities" section right above it that relies on recentActivities.length === 0, and it's supposed to be an either/or situation: either the table or no recent activities. For that reason, it's much cleaner to rely on the same property, as opposed to two parallel properties. It's a small nuance, but I do think it's an improvement. This checks the actual data length rather than a value that's fed to us.

src/app/project/project-activities/project-activities.component.ts
jareth-whitney: It's really 6 of one, 1/2 a dozen of the other. I can certainly do that if you think it matters.

jareth-whitney: And yes on the type safety

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: Technically it's the project and the banner, so it would be inaccurate to say it's just the project. I'll still rename it to make it more clear though.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: None of the subscribe responses in this app are typed, I don't think that's necessary.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: I added a doc block here. I added the type as well, but it's literally just 'Data'.

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: "this" isn't available within the "onAdd" so you have to reassign before usage in that context. L had to be reassigned to this.L to trick the compiler into not auto-injecting Leaflet.

src/app/services/api.ts
jareth-whitney: It stops API calls with undefined values

src/app/services/scriptLoader.service.ts
jareth-whitney: I'll make it clear that the use case is to bypass app-level script loading.

src/app/services/scriptLoader.service.ts
jareth-whitney: That wouldn't work here, since you're just returning promises. You would have to catch the errors on the receiving end of the data.

CHANGELOG.md
jareth-whitney: I took out all of the nested lines. I don't think it's bad to have the extra detail but I digress.

src/app/leaflet/leaflet-augmentations.ts
jareth-whitney: I'll just call it leaflet declarations and leave some more context for people.

src/app/projects/projlist-map/projlist-map.component.ts
jareth-whitney: No it's not the same at all. Lazy loading diverts the loading of the Angular component but you would still have to bundle the script, so it would be loaded into one of the main bundle files. This completely avoids that.

src/app/projects/projlist-map/projlist-map.component.ts
jareth-whitney: The likelihood is low but I'll still add it to be safe.

src/app/projects/projlist-map/projlist-map.component.ts
jareth-whitney: Best to replace it even here, to avoid accidental script bundling by the main app, in the event that someone uncomments it.

src/app/project/project-activities/project-activities.component.ts
Baelx: I don't think leaving it in place and not leaving it in place are equivalent. The way it looks is that you've changed the logic so that it gets project data from another source, but you've left in place part of the logic that gets project data from the resolver. You could also just rename "project" to "currentProject" if you wanted to save some work. But I don't see why we need to have this extra step of assigning `this.project` to `this.currenctProject` other than it's fastest to implement.

src/app/project/project-details-tab/project-details-tab.component.ts
Baelx: It's true that many aren't typed but that doesn't mean we should continue that pattern. If you're adding or restructuring a bunch of code, it should all be typed unless there's a really good reason for it not to be. There may be one example of a subscribe handler being typed in LUP public, if it helps to see how it's done

src/app/project/project-details-tab/project-details-tab.component.ts
Baelx: I'm still not seeing a doc block in the code for some reason. Am I missing it? Could you provide me with a link to see it?

src/app/project/project-details-tab/project-details-tab.component.ts
Baelx: I see, thanks for explaining. I'm not a fan of how this was originally set up (not by you) to use the "this" keyword but it is what it is

src/app/projects/projlist-map/projlist-map.component.ts
Baelx: It feels weird to maintain commented out code that's not even used. Please just remove it

src/app/services/scriptLoader.service.ts
Baelx: Sure fair enough, maybe adding the catch handler down the chain would be better for your code. Either way, I was just hoping to see some more error handling for the end-user

src/app/project/project-details-tab/project-details-tab.component.ts
Baelx: nevermind, found it!

Baelx:
Still looking for a few things in this PR before we proceed:
- Please remove any commented-out code you may have modified. I can tell you right now that if you see any commented-out code in LUP that doesn't have an explanation, it can be removed. Our code commenting policy is that you should strive to never do it, but if you must comment-out code, it needs to have an explanation for why it was done
- The app isn't perfect in terms of its typing but whenever we add or refactor code, we should be typing everything we can or commenting as to why we can't
- If the loadScripts/Styles functions have an error, we should notify the user that something isn't working and redirect them to the home page. Also, I've done some local testing, and I don't think the error in `reject` gets bubbled up to the try/catch block. You may need to refactor a bit there: https://javascript.info/promise-error-handling

src/app/project/project-activities/project-activities.component.ts
jareth-whitney: I get why the hierarchy of this situation is a bit confusing. It was hard to implement everything. This is the structure that this component is using:

project > project-details > project-activities. So project details is loaded within project, and project-activities is loaded within project-details. When I load project-details, I run the activities resolver for project-activities and then pass the value down to that component as an @input. For the project specifically, I am grabbing that from the parent data snapshot in project-details (which uses our original project resolver, no duplicate calls), and then that data is passed down to project-activities through an @input.

When you have an input like project, it's assigned to this.project. I changed it from using this.currentProject to this.project in my last commit, which is why I marked this conversation resolved. It is just property shuffling though, hence the comment. I also wrapped any calls to the project._id with conditionals. So what is the problem now?

src/app/project/project-details-tab/project-details-tab.component.ts
jareth-whitney: I get what you're saying, and I'm not opposed to going back and typing all of the subscribes in the app, but I don't think this is the forum.

src/app/projects/projlist-map/projlist-map.component.ts
jareth-whitney: Sure, I'll remove it.

src/app/services/scriptLoader.service.ts
jareth-whitney: I did add some try/catch blocks

src/app/models/recentActivity.ts
```diff
@@ -0,0 +1,25 @@
+export class RecentActivity {
```
Baelx, line 1: Thanks for adding in this model for RecentActivity! Maybe we can get more use out of it in the frontend going forward

src/app/projects/projlist-map/projlist-map.component.ts
```diff
@@ -81,7 +85,9 @@ export class ProjlistMapComponent implements AfterViewInit, OnChanges, OnDestroy
         '/leaflet-markercluster/leaflet.markercluster.js',
       ]);
     } catch (e) {
-      console.error('One or more scripts failed to load', e)
+      console.error('One or more scripts failed to load', e);
+      alert('Uh-oh, the map failed to load. You will be redirected to the homepage.');
```
Baelx, line 89: Thanks for helping the user out in the event of an error!

Baelx (approved):
I appreciate the updates! Approved

PR #175, opened 2025-08-09
Feature/deseng865: Fixed small footer on map page, improved pop-up behaviour, adjusted map height for different viewports

jareth-whitney:
- Fixed small footer on map page
- improved pop-up behaviour
- adjusted map height for different viewports
- Ticket: https://citz-gdx.atlassian.net/browse/DESENG-865

### DIFF
--- a/src/app/footer/footer.component.html
+++ b/src/app/footer/footer.component.html
@@ -1,13 +1,27 @@
-<footer class="app-footer--sm" *ngIf="router.url.split('#')[0] === ('/projects') || router.url.split(';')[0] === ('/projects')">
+<footer class="app-footer--sm" [class.hidden]="footerHidden" *ngIf="router.url.split('#')[0] === ('/projects') || router.url.split(';')[0] === ('/projects')">
   <div class="container">
-    <div class="footer-admin clearfix">
+    <div class="footer-admin">
       <ul class="gov-links">
         <li><a routerLink="/home">Home</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/copyright" target="_blank">Copyright</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/disclaimer" target="_blank">Disclaimer</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/privacy" target="_blank">Privacy</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/accessibility" target="_blank">Accessibility</a></li>
       </ul>
+      <div 
+        class="close-footer" 
+        [attr.aria-hidden]="!isMobile ? 'true' : null"
+        [attr.aria-label]="isMobile ? 'Show or hide the footer' : null" 
+        (click)="showHideFooter()"
+      >
+        <span>
+          {{ footerHideSymbol }}
+        </span>
+      </div>
     </div>
   </div>
 </footer>
--- a/src/app/footer/footer.component.html
+++ b/src/app/footer/footer.component.html
@@ -1,4 +1,4 @@
-<footer class="app-footer--sm" [class.hidden]="footerHidden" *ngIf="router.url.split('#')[0] === ('/projects') || router.url.split(';')[0] === ('/projects')">
+<footer class="app-footer--sm" [class.hidden]="isFooterHidden" *ngIf="router.url.split('#')[0] === ('/projects') || router.url.split(';')[0] === ('/projects')">
   <div class="container">
     <div class="footer-admin">
       <ul class="gov-links">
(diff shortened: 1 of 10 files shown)


src/app/footer/footer.component.html
```diff
@@ -1,13 +1,27 @@
...
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/copyright" target="_blank">Copyright</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/disclaimer" target="_blank">Disclaimer</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/privacy" target="_blank">Privacy</a></li>
+        <li aria-hidden="true">|</li>
         <li><a href="https://www2.gov.bc.ca/gov/content/home/accessibility" target="_blank">Accessibility</a></li>
       </ul>
+      <div
```
Baelx: Remember to make markup semantic whenever you can! This should be a button

https://www.w3schools.com/html/html5_semantic_elements.asp

src/app/footer/footer.component.scss
```diff
@@ -73,41 +73,82 @@
...
   bottom: 0;
   left: 0;
+  width: 25%;
   border: none;
   position: absolute;
   color: #fff;
   font-size: 0.65rem;
+  background-color: grey;
+  display: flex;
+  z-index: 200000;
```
Baelx: In general, when using an SPA, there should be less of a need to do old-school browser tricks like setting the z-index arbitrarily high so the UI looks right. Maybe when the footer is shown, you can hide the acknowledgements? Or you could shrink the map container so the footer can fit without blocking anything

src/app/footer/footer.component.ts
```diff
@@ -1,12 +1,33 @@
...
-  constructor(public api: ApiService, public router: Router) {}
+  constructor(public api: ApiService, public router: Router, private breakpointObserver: BreakpointObserver) {
+
+    // Get the viewport to determine how we display the footer show/hide toggle.
+    this.breakpointObserver.observe(['(max-width: 767px)'])
+      .subscribe(result => {
+        this.isMobile = result.matches;
+      });
+  }
+  public footerHidden = false;
```
Baelx: It'd be nice to standardize the naming of the booleans here: `isFooterHidden`

src/app/footer/footer.component.ts
```diff
@@ -1,12 +1,33 @@
...
+  public footerHideSymbol = '<';
+  public isMobile: boolean;
+
+  /**
+   * Shows or hides the footer in mobile mode
+   * 
+   * @returns void
+   */
+  public showHideFooter = function() {
```
Baelx: FYI, in the context of a class, this creates a property and binds a function to it. Ideally this would be a method. To implement a method, just remove the assignment operator and the function keyword.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

src/app/footer/footer.component.scss
```diff
@@ -73,41 +73,82 @@
...
+  min-height: 65px;
+
+  &.hidden {
+
+    @media (max-width: 768px) {
+      width: 30px;
+    }
+
+    div ul.gov-links.gov-links.gov-links {
+      display: none;
```
Baelx, line 93: You don't necessarily have to change this but going forward, you should aim to remove DOM elements the Angular way, with `*ngIf*. Single Page Applications were conceived of, in part, to move beyond the need for CSS-based UI tricks and put Javascript more in control of the UI. In hiding a component or element the Angular way, you save some memory too:

https://stackoverflow.com/questions/50558124/whats-the-difference-between-displaynone-and-ngif-false

Baelx (changes requested):
Thanks for this! Please have a look at my comments but to proceed I'll just ask that you switch the div to a button for the footer toggle.

src/app/footer/footer.component.scss
jareth-whitney: It's not a trick, it just needs to be rendered in front of the map. I just inspected the map and its z-index is 999, so I'll give the footer a z-index of 1000 and leave a comment about the map z-index.

src/app/footer/footer.component.ts
jareth-whitney: That's just muscle memory, fixed.

src/app/footer/footer.component.scss
jareth-whitney: No worries, I changed it to *ngIf for the button. The other part I still need to slide off screen though. I tried just doing an ngIf for the ul and it didn't look very clean.

jareth-whitney: The up side to hiding the footer instead of removing it from the DOM is that it can still be read by screen readers.

src/app/footer/footer.component.scss
jareth-whitney: map filters are 1004 so I ended up making it 1005, and I included the values of both the map and the project filters in the comment.

Baelx (approved):
Thanks for adding the button and comments to the z-index declaration!

PR #176, opened 2025-10-28
DESENG-899: Swap position of comment period dates and description

NatSquared:


jareth-whitney (approved):
Thanks for switching the description and date. No individual comments were necessary.

PR #178, opened 2025-10-28
DESENG-901: Add PDF link to homepage

NatSquared:
- Update homepage to link the text "land and water planning in B.C." to the official gov't resource PDF. [DESENG-901](https://citz-gdx.atlassian.net/browse/DESENG-901)
  - Added appropriate aria-labels and icons to indicate link opens in new tab and is a PDF.


src/app/home/home.component.html
```diff
@@ -82,50 +83,60 @@ <h2>About this Site</h2>
...
-               The Province is adopting a more integrated approach to planning that will result in
-               coordinated processes to achieve various planning objectives. This site provides access
-                to information on all active land use planning, forest landscape planning, and water
-                 planning projects.
+          <a href="https://www2.gov.bc.ca/assets/gov/farming-natural-resources-and-industry/forestry/forest-landscape-plans/land-and-water-planning.pdf"
+            target="_blank" rel="noopener noreferrer"
+            aria-label="Land and water planning in B.C. (PDF, 2.6 MB, opens in new tab)">
+            Land and water planning in B.C.
+            <i class="material-icons" aria-hidden="true">picture_as_pdf</i>
+            <i class="material-icons" aria-hidden="true">open_in_new</i>
```
jareth-whitney, line 94: I think we decided against using these icons on the site. I was also an advocate of them. I won't block the PR, but it might be worth checking with Steve.

jareth-whitney (approved):
Looks good, thanks for adding the PDF link. Check with Steve regarding open_in_new icon, as I was previously asked not to use it.

PR #180, opened 2025-10-29
DESENG-882: Fix unit tests & linting issues

NatSquared:
- Stabilized Angular unit tests by stubbing heavy components and relaxing template validation with `NO_ERRORS_SCHEMA`  so `ng test` runs again. [DESENG-882](https://citz-gdx.atlassian.net/browse/DESENG-882)
  - Replaced conflicting third-party typings with local shims (for example Leaflet marker clusters).
  - Brought remaining spec imports, forms modules, and router stubs up to date to eliminate CI failures caused by missing declarations or runtime errors.
  - Added skipLibCheck to tsconfig.spec.json to bypass type checking issues with Karma and Jasmine.


src/app/services/commentperiod.service.spec.ts
```diff
@@ -49,51 +48,41 @@ describe('CommentPeriodService', () => {
...
 
         service
           .getAllByProjectId('123')
           .subscribe(result =>
-            expect(result).toEqual([
-              new CommentPeriod({ _id: '1' }),
-              new CommentPeriod({ _id: '2' }),
-              new CommentPeriod({ _id: '3' })
-            ])
+            expect(result).toEqual({
```
jareth-whitney, line 68: Wow, this one must have been old if totalCount wasn't factored in.

src/tsconfig.spec.json
```diff
@@ -5,6 +5,7 @@
     "module": "commonjs",
     "target": "es5",
     "baseUrl": "",
+    "skipLibCheck": true, // bypass Karma type checking issues
```
jareth-whitney: Not a huge deal but this line was highlighted in bright red in Github, I think because JSON usually doesn't allow comments, but the compiler should be able to parse this just fine.

jareth-whitney (approved):
Thanks for taking the time to revive all of these old unit tests. This will be much quicker than re-writing our unit tests from scratch. Assuming the test suite is completing with 0 errors on your end, I don't see anything blocking this PR. You can optionally remove your comment from tsconfig.spec.json, but I'm just splitting hairs.

PR #184, opened 2025-11-15
Feature/deseng914: Added Scully pre-rendering for static pages.

jareth-whitney:
- Added pre-rendering for static pages with Scully [DESENG-914](https://citz-gdx.atlassian.net/browse/DESENG-914)
  - Customized package.json to work with custom Scully/Puppeteer/Chromium configuration compatible with ARM macs and TS 4.0.8
  - Modified dist folder layout and app name to match Scully's requirements
    - Updated all references
  - Updated OpenShift pipeline and buildConfigs to accomodate pre-rendering stage

** NOTE: OpenShift buildConfigs and pipeline will be updated in OC Console once I receive approval on this PR **


e2e/app.po.ts
```diff
@@ -1,6 +1,6 @@
 import { browser, element, by } from 'protractor';
 
-export class BaseApp2Page {
+export class LandUsePlanningPublicPage {
```
NatSquared, line 3: Much more descriptive!

e2e/app.e2e-spec.ts
```diff
@@ -1,14 +1,15 @@
-import { BaseApp2Page } from './app.po';
+import { LandUsePlanningPublicPage } from './app.po';
 
-describe('base-app2 App', () => {
-  let page: BaseApp2Page;
+describe('Land Use Planning Public App', () => {
+  let page: LandUsePlanningPublicPage;
```
NatSquared, line 4: Thanks for updating these old tests!

package.json
```diff
@@ -1,19 +1,26 @@
...
+    "postbuild": "scully",
     "pretest": "",
     "test": "echo 'Test suite should be run with \"`npm run tests` or npm `run tests-ci`\"'",
     "tests": "ng test",
     "tests-ci": "ng test --watch=false",
     "lint": "ng lint --format=stylish",
     "e2e": "ng e2e",
-    "postinstall": "ng build --prod --configuration=production"
+    "preinstall": "export PUPPETEER_SKIP_DOWNLOAD=true",
+    "postinstall": "unset PUPPETEER_SKIP_DOWNLOAD && npm run build",
```
NatSquared: Nice workaround! I assume this means you have to take care of installing the browser yourself? Where does it look for the binary?

src/assets/scully-routes.json
```diff
@@ -0,0 +1 @@
+[{"route":"/lup"},{"route":"/wsp"},{"route":"/flp"},{"route":"/faq"},{"route":"/contact"},{"route":"/compliance-oversight"}]
```
NatSquared, line 1: We should also prerender the home page (`/`), unless there's dynamic content there I'm not seeing
```suggestion
[{"route":"/"},{"route":"/lup"},{"route":"/wsp"},{"route":"/flp"},{"route":"/faq"},{"route":"/contact"},{"route":"/compliance-oversight"}]
```

src/polyfills.ts
```diff
@@ -96,3 +96,9 @@ if (!String.prototype.padStart) {
     }
   };
 }
+
+/***************************************************************************************************
+* SCULLY IMPORTS
+*/
+// tslint:disable-next-line: align
+import 'zone.js/dist/task-tracking';
```
NatSquared: Does this have a performance impact outside of a Puppeteered browser?

.gitignore
```diff
@@ -36,6 +36,11 @@ yarn-error.log
 yarn.lock
 debug.log
 
+# Pre-rendering
+scully.log
```
NatSquared, line 40: Being a good steward of the file tree and keeping it clean!

angular.json
```diff
@@ -3,15 +3,15 @@
   "version": 1,
   "newProjectRoot": "projects",
   "projects": {
-    "base-app2": {
+    "landuseplanning-public": {
```
NatSquared, line 6: Glad these "base app" references are getting renamed to something descriptive :)

NatSquared (approved):
Looks to my eye like a textbook Scully implementation. Just a few comments to review, including taking advantage of rendering the homepage statically if that's an option. Have you tested with the recent SEO changes to make sure OpenGraph tags still appear on statically rendered pages?
Assuming all is good on those fronts, approved!

package.json
jareth-whitney: Thank you for pointing this code out to me. I believe I can remove these env variables now. They were originally implemented because Scully 1.1.1 has an internal dependency of Puppeteer 5.5.0. That version of Puppeteer needs a specific Chromium browser build, which it tries to install. That fails, because the version of Chromium isn't compatible with M2 macs. But now that I'm using overrides instead of resolutions, Scully 1.1.1 just references Puppeteer 10.4.0, which subsequently downloads a newer Chromium build that does work with M2 Macs.

I will test a full reinstall without the env variables in the pre/post install.

Edit: Confirmed working without these variables. They have been removed.

src/assets/scully-routes.json
jareth-whitney: As discussed, it's unfortunately dynamic right now. But we should have a ticket for removing the updates section and converting it to static, pending Steve's approval.

src/polyfills.ts
jareth-whitney: It does have a small performance impact, good catch. Adding this import to the polyfills.ts was part of the Scully init script. I added a conditional around it, so it only runs for Scully purposes now:


if (typeof window === 'undefined') {
  require('zone.js/dist/task-tracking');
}

PR #185, opened 2025-11-19
Feature/deseng914: Made updates to Scully pipeline integration

jareth-whitney:
- Updated build config for pre-rendering
- moved Scully to prod dependencies
- added image stream yaml
- updated pipeline yaml
- modified angular-builder Dockerfile


openshift/templates/angular-builder/Dockerfile
```diff
@@ -1,46 +1,12 @@
-# Get community edition of nodejs v10.x (LTS)
-FROM centos/nodejs-10-centos7
-
-ENV PATH="/opt/rh/rh-nodejs10/root/usr/bin/:$PATH"
+# Get nodejs version 16 ubi8 from local OpenShift image registry
+FROM image-registry.apps.silver.devops.gov.bc.ca/openshift/nodejs:16-ubi8
```
NatSquared, line 2: I assume upgrading the Node version is backwards compatible in this repo?

package.json
```diff
@@ -38,6 +38,7 @@
     "@angular/router": "^11.2.14",
     "@bcgov/bc-sans": "^2.0.0",
     "@ng-bootstrap/ng-bootstrap": "^9.1.3",
+    "@scullyio/scully": "^1.1.1",
```
NatSquared, line 41: Yep; this makes more sense to me as a regular dependency 😆

openshift/templates/pipeline/lup-public-build-deploy-dev.yaml
```diff
@@ -69,7 +69,7 @@ spec:
             oc start-build lup-public-angular-builder --follow --wait
       taskRef:
         kind: ClusterTask
-        name: openshift-client
+        name: openshift-client-1-5-0
```
NatSquared, line 72: What does "1-5-0" refer to here?

openshift/templates/buildConfigs/lup-public-prerender-static-pages
```diff
@@ -32,16 +32,13 @@ spec:
...
-          yum install -y nodejs git && \
-          npm install -g npm@8
-
       WORKDIR /opt/app-root/src
 
-      RUN npm ci && \
-          npx scully --config scully.config.js
+      COPY --from=image-registry.openshift-image-registry.svc:5000/e8b9ad-tools/lup-public-compiled:latest /opt/app-root/src/dist/landuseplanning-public ./dist/landuseplanning-public
+      COPY --from=image-registry.openshift-image-registry.svc:5000/e8b9ad-tools/lup-public-compiled:latest /opt/app-root/src/package*.json ./
+      COPY --from=image-registry.openshift-image-registry.svc:5000/e8b9ad-tools/lup-public-compiled:latest /opt/app-root/src/scully.landuseplanning-public.config.ts ./scully.landuseplanning-public.config.ts
```
NatSquared, line 39: Huh nice; I didn't know you could copy out of other images in Dockerfiles like that

NatSquared (approved):
Looks good to me:
- Moving to an internal nodejs image with a later version (10 -> 16)
- Changing docker workflows to copy files from the compiled LUP page instead of using it directly as a base
- Moved Scully from devDependencies to dependencies

openshift/templates/pipeline/lup-public-build-deploy-dev.yaml
jareth-whitney: It refers to version 1.5.0. The pipeline was failing because it couldn't identify the ClusterTask with just "openshift-client". I'm not sure why this has changed.

package.json
jareth-whitney: Yeah exactly, hindsight is 20/20.

openshift/templates/buildConfigs/lup-public-prerender-static-pages
jareth-whitney: I believe I have access because I defined the imagestream in the Dockerfile, but the main objective here is to avoid compiling the Angular twice, which would slow down the pipeline quite a bit. I'm trying to take the Angular compile from the previous pipeline step, copy the files, and run Scully on top of that, then deploy the Scully imagestream that results.

openshift/templates/angular-builder/Dockerfile
jareth-whitney: I deleted node_modules and package-lock.json locally then did a full install with Node 16.16 and NPM 8 with no problems resulting. That's not a guarantee of compatibility, but it's a good first step. If it doesn't work then I will try rolling back to Node 14 and install NPM 8 manually in this Dockerfile.

PR #186, opened 2025-11-20
Feature/deseng914: Further refinements to the Scully pre-rendering flow

jareth-whitney:
- Updated build config for pre-rendering
- Moved Scully to prod dependencies
- Added image stream yaml
- Updated pipeline yaml
- Modified angular-builder Dockerfile


openshift/templates/buildConfigs/lup-public-compiled.yaml
```diff
@@ -0,0 +1,35 @@
...
+spec:
+  nodeSelector: null
+  output:
+    to:
+      kind: ImageStreamTag
+      name: 'lup-public-compiled:latest'
+  resources:
+    limits:
+      cpu: 500m
+      memory: 4Gi
```
NatSquared, line 16: This'll be enough for Scully and the Puppeteer instance?

openshift/templates/buildConfigs/lup-public-deploy-image
```diff
@@ -33,18 +33,16 @@ spec:
   source:
     type: Dockerfile
     dockerfile: |-
-      FROM lup-public-nginx-runtime:latest
```
NatSquared, line 36: The FROM directive is still needed here in order for Openshift to parse the Dockerfile correctly: 
```suggestion
      FROM lup-public-compiled:latest
```

NatSquared (approved):
Further work towards integrating Scully into the pipeline; looks good!
Please just have a look at my comment about deleting the FROM statement in an embedded Dockerfile; that statement is still required

New approach: 
- Removed separate prerender step and use postbuild script to trigger it in the lup-public-compiled container
- Added dependencies for puppeteer 
- Change lup-public-deploy build strategy to copy files post-execution

openshift/templates/buildConfigs/lup-public-compiled.yaml
jareth-whitney: I believe that's 100% of what we've been allotted. Maybe the serverlords will grant us another CPU core.

PR #190, opened 2025-12-22
Always display environment banner in public view

NatSquared:
[DESENG-924](https://citz-gdx.atlassian.net/browse/DESENG-924)
Update environemnt banner component to read from correct localStorage key, fixing an issue where the banner would not display when it should unless the current user had a localStorage entry from the admin-facing server.

### DIFF
--- a/src/app/header/env-banner/env-banner.component.ts
+++ b/src/app/header/env-banner/env-banner.component.ts
@@ -12,7 +12,7 @@ export class EnvBannerComponent implements OnInit {
   public showBanner: boolean;
 
   constructor() {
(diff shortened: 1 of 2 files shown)


src/app/header/env-banner/env-banner.component.ts
```diff
@@ -12,7 +12,7 @@ export class EnvBannerComponent implements OnInit {
   public showBanner: boolean;
 
   constructor() {
-    const deployment_env = window.localStorage.getItem('from_admin_server--deployment_env');
```
jareth-whitney, line 15: Nice quick win.

jareth-whitney (approved):
Crazy how this one flew under the radar for so long! Thanks for fixing it.

PR #195, opened 2026-04-29
bugfix/deseng959: Refactored participate code, cleaned up imports

jareth-whitney:
- Refactored participate now code into injectable for less repeated code [DESENG-959](https://citz-gdx.atlassian.net/browse/DESENG-959)
  - Created a participate service that fetches modal with the correct data
- Created unit tests for new service
- Removed unused imports from multiple components


NatSquared (approved):
Thanks for making these changes! With this more centralized code, it should be easier to modify the behaviour across pages in the future

